PR #68, opened 2021-10-29
Refactor Auction Structure

tbrent:
- This PR reduces the amount of auction-specific logic to a minimum, making use of one-size-fits-all helper functions to prepare auctions.
- This approach enables a more robust approach to revenue auctions. Previously we were counting on being able to forget about the revenue tokens after launching our revenue auction. This wouldn't work well in situations where the revenue auction fails to clear, though.


julianmrodri (approved):
LGTM MErging and will adjust tests.

PR #69, opened 2021-11-01
Fix tests

julianmrodri:
* Fix tests and minor changes for `Vault`, `Asset` and `stRSR`.


contracts/proto0/StRSRP0.sol
```diff
@@ -141,7 +141,11 @@ contract StRSRP0 is IStRSR, Context {
         processWithdrawals();
 
         uint256 snapshotTotalStakedPlus = _totalStaked + _amountBeingWithdrawn();
-        _totalStaked -= amount;
+
+        // Only decrease if there are enough Staked funds
+        if (_totalStaked >= amount) {
```
tbrent, line 146: Don't we want the tx to revert if more than the total staked is seized?

julianmrodri: This is a valid scenario where all funds are pending `withrawal`. You can have 0 stake and amounts that can still be seized.

PR #76, opened 2021-11-05
Main and Oracle test examples

julianmrodri:
* `Main` (issuance and slow minting) initial tests. They are based on the previous version before `Fix` and also include changes to contracts which are being modified.

* Also includes an `Oracle` test example that relies on Mainnet Forking (tests are commented with `describe.skip`) Only should be run with appropiate Alchemy configuration.
 
**This PR is not intended for merge but to use as temporary reference/example**


tbrent:
Sweet, I've mined this for changes to `fixed-point-integration`. Closing

PR #82, opened 2021-11-12
feat: proto0 deployment scripts

lcamargof:
## Description of changes
- Proto0 deployments scripts

## Note
It requires `allowUnlimitedContractSize: true` as the RToken contracts are too large.


tbrent:
LGTM

PR #127, opened 2021-12-27
fix time logic

julianmrodri:
* Fixed logic broken by a previous refactor. Tests now passing again for StRSR
Note: Confirm the `>=` if it seems correct or we want exclusive `>`


tbrent:
LGTM

PR #133, opened 2022-01-10
 Add vault parameter to withdrawBUs

julianmrodri:
* Adds the vault parameter to `withdrawBUs` so it can redeem from the old vault


tbrent:
LGTM

PR #134, opened 2022-01-11
p0-nits Unit of Account

tbrent:


contracts/p0/assets/Collateral.sol
```diff
@@ -63,8 +60,7 @@ contract CollateralP0 is ICollateral, AssetP0, Context {
 
         // If the underlying fiatcoin price is below the default-threshold price, default eventually
         if (whenDefault > block.timestamp) {
-            Price memory p = fiatcoinPrice(); // {attoPrice/fiatTok}
-            whenDefault = p.quantity(uoa).lte(_defaultThreshold())
+            whenDefault = fiatcoinPrice().lte(main.defaultThreshold())
```
fiddlemath, line 63: 🤔  -- bad smell here; we'd previously been treating main.defaultThreshold as a ratio, not a price. Maybe we need `function minFiatcoinPrice()`, which for our current assets is just `return main.defaultThreshold();`?

contracts/p0/libraries/Rewards.sol
```diff
@@ -26,21 +26,23 @@ library RewardsLib {
...
+        //     // `collateral` being unused here is expected
+        //     // compound groups all rewards automatically
+        //     main.oracle(UoA.USD).comptroller().claimComp(address(this));
+        //     sweepAmt = main.compAsset().erc20().balanceOf(address(this));
+        //     if (sweepAmt > 0) {
+        //         main.compAsset().erc20().safeTransfer(address(main), sweepAmt);
+        //     }
+        // }
+        return 0;
     }
```
fiddlemath, line 47: Finish with and eliminate commented-out code.

contracts/p0/main/Auctioneer.sol
```diff
@@ -99,6 +99,16 @@ contract AuctioneerP0 is
...
+    function owner()
+        public
+        view
+        virtual
+        override(Ownable, SettingsHandlerP0, VaultHandlerP0)
+        returns (address)
+    {
+        return super.owner();
+    }
+
```
fiddlemath, line 111: Oooof - I still don't see why we should thread this through all the contracts -- we _don't_ want to change the idea or behavior of `owner` access between them, right?

fiddlemath (changes requested):

contracts/p0/assets/Collateral.sol
tbrent: Good catch here, was going to fast and didn't catch the units. Applying your suggested fix

PR #151, opened 2022-02-11
generalize reward assets + tie to ClaimAdapters

tbrent:


julianmrodri:
Nice!... Maybe we can emit on `RewardClaimed` two params,  the address of the Reward token and the claimed amount? Whatdo you think?

tbrent:
@julianmrodri added -- I didn't get to updating the tests that were asserting `RewardClaimed` was emitted.

PR #155, opened 2022-02-17
basket assets, do not merge yet

tbrent:
I started out trying to see if I could easily merge the AssetRegistry and BasketHandler. This turned out to not be the correct move, but in getting to that point I noticed lots of ways things weren't as simple as they could be. This PR is really two somewhat coupled changes together: (i) Mixin relationship changes, and (ii) basket configuration/switching. 

Proposed changes:
(i) **Mixin relationships** 
- Move SettingsHandler from being asset-aware to asset-unaware, making it the most base mix-in (which feels right; protocol settings are most-meta)
- Make the sole job of AssetRegistry to provision a unique `ERC20 -> Asset` lookup for the rest of the system to use. It provides functions for registering and unregistering assets, and reverts in order to ensure the caller knows when they are replacing an ERC20 vs not. 
- Switch traders to targeting an ERC20 rather than an asset. When they execute their auction loops, they fetch the latest asset for that ERC20 from Main. 
- Move basketLib functions into internal views that the BasketHandler provides on top of a private basket var, rather than accessing the basket from child classes. Previously the Auctioneer was sharing access to `basket` with the BasketHandler. 

(ii) **Basket Setting**
I'll preface this with the caveat that I could be failing to remember a reason why things were the way they were, so maybe we backpedal on this part of the PR. I currently think it's a good direction though, and I think it's worth considering this simplification.

I claim we have 3 baskets: the collateral basket, the reference basket, and the target basket. 
A. The collateral basket is an ephemeral view on top of the reference basket that changes block-to-block.  
B. The reference basket is a relatively long-lived point in the _space of all possible baskets_ described by the target basket. 
C. The target basket is the north star. It describes what a basket unit is _with respect to_ on a long timescale, where the language it uses to express this consists of amounts of a target unit (typically USD, EUR, etc) and the collateral that can reasonably be used to track those targets.  

Said another way, the reference basket is a straightforward function of the target basket plus the asset registration status plus the collateral default status. You could even imagine never storing the reference basket and simply constructing it dynamically like we do the collateral basket. I think we shouldn't do this because of gas costs, but it's conceptually viable. 

How reference basket setting works:
- For each target unit, iterate through the backup collateral, using each collateral up to its maxAmt, and only if the collateral is SOUND and in the asset registry. 
- After: scale weights uniformly upwards if the collateral maxAmts were not sufficient to cover the target amount.  

How swapping out an asset (with a shared ERC20) works: 
- call `BasketHandler.setTarget()`, setting a new list of collateral for the relevant target. Place the new asset adjacent to its ERC20-sibling, on whichever side. 
- call `AssetRegistry.swapRegisteredAsset(newAsset)`, unregistering the old asset and registering the new one. Immediately this will make the current reference basket invalid, causing the next setting of the reference basket to contain the newAsset in place of the oldAsset

Tests have not been updated yet


tbrent:
Noting that Matt talked me out of the Target Basket version of things. Prime Basket is right.

PR #167, opened 2022-02-25
sample implementation of a generic SettingsHandler

tbrent:
We'll leave this here for a short time in case we decide to roll our own solution for parameter dynamism, but the expectation is for macros and pre-processing to be a better solution.


tbrent:
tagged as `settings-handler-homegrown-generification` and branch deleted

PR #187, opened 2022-03-22
Ensure strong basket invariant of no defaulted or unregistered collateral

tbrent:
The BasketHandler guarantees two invariants: (i) the reference basket should not contain defaulted or unregistered collateral, and (ii) the reference basket is either worth the number of target units in the prime basket, or it is defaulted. 

Changes:
- Do not leave bad collateral in reference basket; remove as soon as collateral becomes bad
- Emit `BasketSet` whenever `_switchBasket` happens, including default status
- Exclude defaulted ERC20s from all BasketHandler views
- Prevent BackingManager from trading if the basket is defaulted or iffy (TO CHECK: Do we agree that it shouldn't trade when iffy?)
- minor and unrelated: I noticed we could eliminate `maxIssuable` from our system and move it out to the facade, so I did that

### DIFF
--- a/contracts/interfaces/IBasketHandler.sol
+++ b/contracts/interfaces/IBasketHandler.sol
@@ -9,11 +9,9 @@ import "./IComponent.sol";
 /**
  * @title IBasketHandler
  * @notice The BasketHandler aims to maintain a reference basket of constant target unit amounts.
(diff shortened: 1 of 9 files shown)


fiddlemath (approved):
👍

PR #192, opened 2022-03-29
StRSR extreme bounds tests

tbrent:
tests for extreme values in StRSR

introduces "Turbo Mode", which is enabled by default in the majority of our test targets. When set to false, a larger branching factor is applied to the test, testing ~2100 cases instead of 128


tbrent:
I want to make sure the workflow cloud tests don't break on anything I introduced, but otherwise this is ready for merge.

PR #195, opened 2022-04-01
Unified Initialization

julianmrodri:
* Adapts P0 and P1 to only use `initializers` (single strategy using OZ Initializer for all)
* Constructor Arguments are removed and contracts only receive the required params for initialization
* All initializations are called from Deployer and not through Main
* Use upgradeable version of contracts for all components when required
* Adapts also the fuzzing contracts to the new initialization strategy
* Adapts tests and new checks

All tests passing P0-P1


contracts/p0/Main.sol
```diff
@@ -21,23 +25,23 @@ contract Pausable is Ownable, IPausable {
         _;
     }
 
-    function pause() external {
```
tbrent, line 24: Is there a reason these had to become public?

julianmrodri: Finally I used external for all and set the Owner and Pauser from the deployer as agreed.

tbrent:
LGTM!

PR #196, opened 2022-04-06
Feat - Upgradeability

julianmrodri:
* Bump and setup OZ Upgradeability tools
* Rename `Component` to `ComponentP0` (so we can have a `ComponentP1` - upgradeable)
* Create P1 versions of contracts for upgradeability (remove immutables, using safe Libraries from OZ, etc)
* Adapt tests to be generic and use Interfaces (like `StRSR` and `RToken` did before)
* Move `Pausable` to `mixins` and reuse in both P0/P1
* Add specific tests for Upgradeability (including V2 mocks for each component - only run for P1)
* Documentation explaining the upgrade process at a high level

All tests passing for P0, P1.


tbrent:
huge haul Julian, LGTM

PR #267, opened 2022-06-15
refactor: README.md

DSergiu:
* fixes some typos
* extends repository structure with missing directories
* other changes for consistency reasons


tbrent:
Hi @DSergiu I think I've tracked down what is going on...the way github secrets work is that they are never passed to actions from a fork PR. This makes sense; otherwise they wouldn't be secrets. 

So I think all you need to do in order to get these two PRs to clear our checks is to configure a secret in your fork repo for `ALCHEMY_MAINNET_KEY`. It should just be the key portion of the URL: `https://eth-mainnet.alchemyapi.io/v2/${{ secrets.ALCHEMY_MAINNET_KEY }}`

DSergiu:
> Hi @DSergiu I think I've tracked down what is going on...the way github secrets work is that they are never passed to actions from a fork PR. This makes sense; otherwise they wouldn't be secrets.
> 
> So I think all you need to do in order to get these two PRs to clear our checks is to configure a secret in your fork repo for `ALCHEMY_MAINNET_KEY`. It should just be the key portion of the URL: `https://eth-mainnet.alchemyapi.io/v2/${{ secrets.ALCHEMY_MAINNET_KEY }}`

Something seems off. I do have the repo secret defined but it is not fetched when the pipeline is run.
image

See the env of the [faield run](https://github.com/reserve-protocol/protocol/runs/7258886349?check_suite_focus=true#step:8:6): "MAINNET_RPC_URL: https://eth-mainnet.alchemyapi.io/v2/"
See the env of a [success run](https://github.com/reserve-protocol/protocol/runs/7258170999?check_suite_focus=true#step:8:6): "MAINNET_RPC_URL: https://eth-mainnet.alchemyapi.io/v2/***"

Clearly the secret from the forked repo is not read, mind the missing `***` at the end of the env `MAINNET_RPC_URL`.
Could this be because of some missing setting? @tbrent

tbrent:
@DSergiu I'm accepting #291 but I think we prefer keeping the README as-is. Thanks for taking the time to prepare both these PRs!

PR #279, opened 2022-07-11
Strsr new min exchange rate

fiddlemath:


contracts/p1/StRSR.sol
```diff
@@ -261,46 +262,41 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
...
+        uint256 rewards = rsrRewards();
+
+        // Remove RSR from stakeRSR
+        uint256 stakeRSRToTake = (stakeRSR * rsrAmount + (rsrBalance - 1)) / rsrBalance;
+        stakeRSR -= stakeRSRToTake;
+        seizedRSR = stakeRSRToTake;
+
+        // update stakeRate, possibly beginning a new stake era
+        if (stakeRSR > 0) {
+            stakeRate = uint192((FIX_ONE_256 * totalStakes) / stakeRSR);
```
tbrent, line 275: Isn't it possible for `stakeRSR` to be arbitrarily small, such that this downcast truncates?

PR #336, opened 2022-09-07
better .env.example

jeremyschlatter:
These env vars cause `yarn test:fast` to fail (at least on my machine).

Better to have them unset so new developers (like me) can get the unit tests running faster.


tbrent:
Agreed. Thanks!

PR #342, opened 2022-09-16
Solidified changes

tbrent:


contracts/plugins/assets/EURFiatCollateral.sol
```diff
@@ -73,20 +73,21 @@ contract EURFiatCollateral is Collateral {
...
-
-                // If the price is below the default-threshold price, default eventually
-                if (p < peg - delta || p > peg + delta) {
-                    whenDefault = Math.min(block.timestamp + delayUntilDefault, whenDefault);
-                } else whenDefault = NEVER;
+                    // If the price is below the default-threshold price, default eventually
+                    if (p < peg - delta || p > peg + delta) {
+                        whenDefault = Math.min(block.timestamp + delayUntilDefault, whenDefault);
+                    } else whenDefault = NEVER;
+                }
```
fiddlemath, line 90: What *should* happen if p2 == 0? priceable becomes false, so status() becomes UNPRICED? I think that's probably correct, and what'll actually happen, but it's hard to read that intent off of this code structure.

contracts/plugins/assets/NonFiatCollateral.sol
```diff
@@ -125,11 +125,13 @@ contract NonFiatCollateral is Collateral {
...
     /// @return min {tok} The minimium trade size
     function minTradeSize() external view override returns (uint192 min) {
-        try uoaPerTargetFeed.price_(oracleTimeout) returns (uint192 p) {
+        try uoaPerTargetFeed.price_(oracleTimeout) returns (uint192 p1) {
             try chainlinkFeed.price_(oracleTimeout) returns (uint192 p2) {
                 // {UoA/tok} = {UoA/target} * {target/ref} * {ref/tok}
-                // p = p.mul(p2);
-                p = uint192((uint256(p) * p2) / FIX_ONE);
+
+                // In the next line we violate our standard practice and put D18 on a uint256
```
fiddlemath, line 132: Say not just that we do this, but why it's necessary; otherwise a future developer could sensibly come back through and "clean it up" in the future.

fiddlemath: Ditto on L153.

fiddlemath (approved):

contracts/plugins/assets/EURFiatCollateral.sol
tbrent: Yep, that's intended.

contracts/plugins/assets/NonFiatCollateral.sol
tbrent: Good point.

PR #346, opened 2022-09-19
delete dangling sentence in comment

jeremyschlatter:
I'm not sure how this sentence was meant to complete. Please feel free to counter-propose a completion to the sentence instead.


fiddlemath:
@tbrent do you recall what you intended here?

tbrent:
This is actually out-of-date now, too, from a time when things worked differently. It used to be that `StRSRVotes.getPastVotes` would return 0 if there had been an era change since the voting weight snapshot. We changed this so that the voting weights remain static and put the `Governance` contract in charge of handling era change. 

I think we can delete the whole comment, given that StRSRVotes is no longer surprising in how it handles past voting weight on era change.

jeremyschlatter:
I deleted more of the comment.

PR #350, opened 2022-09-26
Min trade size simplification

tbrent:


contracts/plugins/assets/RTokenAsset.sol
```diff
@@ -4,57 +4,129 @@ pragma solidity 0.8.9;
...
+                uint192 q = shiftl_toFix(
+                    quantities[i],
+                    -int8(IERC20Metadata(erc20s[i]).decimals())
+                );
+
+                // downcast is safe: total attoUoA from any single asset is well under 1e47
+                // D18{UoA} = D18{UoA} + (D18{UoA/tok} * D18{tok} / D18
+                p += uint192((asset.priceWithFailover() * uint256(q)) / FIX_ONE);
+            }
+        }
```
fiddlemath, line 110: This seems fine, but there's heavy repetition of code. This could be two entry points to a single parameterized function, with just a bit like `uint192 assetPrice = failover?asset.priceWithFailover():asset.price()`

fiddlemath: Notably, you don't then need the outermost try-catch in priceWithFailover().

fiddlemath (changes requested):

contracts/p0/mixins/TradingLib.sol
```diff
@@ -23,55 +20,52 @@ uint256 constant GNOSIS_MAX_TOKENS = 7e28;
...
     /// @return notDust True when the trade is larger than the dust amount
     /// @return trade The prepared trade
+    // Recall: struct TradeRequest is {IAsset sell, IAsset buy, uint sellAmount, uint minBuyAmount}
+    //
+    // If notDust is true, then the returned trade satisfies:
+    //   trade.sell == sell and trade.buy == buy,
+    //   trade.minBuyAmount ~= trade.sellAmount * sell.price() / buy.price() * (1-maxTradeSlippage),
+    //   trade.sellAmount <= sell.maxTradeSize().toQTok(sell)
+    //   1 < trade.sellAmount <= GNOSIS_MAX_TOKENS,
+    //   trade.buyAmount <= GNOSIS_MAX_TOKENS,
```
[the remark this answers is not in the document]
tbrent: Yep, good catch.

contracts/plugins/assets/RTokenAsset.sol
tbrent: I'm aware, I tried it both ways. There are 3 differences to hit:
1. The CEIL vs FLOOR difference in the `quote` call
2. The prorata calculation
3. `asset.priceWithFailover()` vs `asset.price()`

which was just enough to get me to revert it to the fully written out thing for both functions.

tbrent: oh, and also a 4th: how `totalSupply == 0` case is treated.

contracts/plugins/assets/RTokenAsset.sol
```diff
@@ -4,57 +4,108 @@ pragma solidity 0.8.9;
...
+
+            if (prorated < quantities[i]) quantities[i] = prorated;
+
+            // D18{tok} = D18 * {qTok} / {qTok/tok}
+            uint192 q = shiftl_toFix(quantities[i], -int8(IERC20Metadata(erc20s[i]).decimals()));
+
+            // downcast is safe: total attoUoA from any single asset is well under 1e47
+            // D18{UoA} = D18{UoA} + (D18{UoA/tok} * D18{tok} / D18
+            p += uint192(
+                (withFailover ? asset.priceWithFailover() : asset.price() * uint256(q)) / FIX_ONE
```
fiddlemath, line 87:
```suggestion
                ((withFailover ? asset.priceWithFailover() : asset.price()) * uint256(q)) / FIX_ONE
```

Right?

contracts/p1/mixins/TradingLib.sol
```diff
@@ -28,13 +28,14 @@ library TradingLibP1 {
     //
     // If notDust is true, then the returned trade satisfies:
     //   trade.sell == sell and trade.buy == buy,
-    //   trade.minBuyAmount ~= trade.sellAmount * sell.price() / buy.price() * (1-maxTradeSlippage),
+    //   trade.minBuyAmount ~= trade.sellAmount * sell.strictPrice() / buy.strictPrice() * (1-maxTradeSlippage),
```
fiddlemath, line 31:
```suggestion
    //   trade.minBuyAmount ~=
    //       trade.sellAmount * sell.strictPrice() / buy.strictPrice() * (1-maxTradeSlippage),
```

contracts/p1/mixins/TradingLib.sol
```diff
@@ -43,18 +44,24 @@ library TradingLibP1 {
...
         uint192 s = fixMin(sellAmount, maxTradeSize(sell));
 
         // {qSellTok}
         trade.sellAmount = s.shiftl_toUint(int8(sell.erc20Decimals()), FLOOR);
 
+        uint192 buyPrice = buy.strictPrice();
+
+        // buy.strictPrice() cannot be zero in BackingManager's usage of this function
+        // buy.strictPrice() could plausibly be zero in RevenueTrader's usage of this function
+        require(buyPrice > 0, "buy asset has zero price");
```
fiddlemath, line 59: So we should put this check in BackingManager itself, right?

contracts/p1/mixins/TradingLib.sol
```diff
@@ -209,30 +225,37 @@ library TradingLibP1 {
         // - Discounting unsound collateral
         // - Discounting dust amounts for collateral in the basket + non-dust assets
 
-        IERC20 rsrERC20 = rsr();
-        IBasketHandler bh = basket();
-        uint192 minTradeVolume_ = minTradeVolume(); // {UoA}
+        IERC20 rsrERC20 = rsr(trader);
+        IERC20 rToken_ = IERC20(address(rToken(trader)));
+        uint192 minTradeVolume_ = trader.minTradeVolume();
```
fiddlemath, line 230:
```suggestion
        uint192 minTradeVolume_ = trader.minTradeVolume(); // {UoA}
```

fiddlemath (changes requested):
I didn't read through p0/.../TradingLib.sol in any detail; but I'm assuming those changes are basically a copy of the p1/ ones.

A few small changes requested here. The thing about saveStatus() could be ignored for now and included in a later PR; other stuff is probably worth dealing with before merging.

contracts/plugins/assets/EURFiatCollateral.sol
```diff
@@ -95,15 +88,17 @@ contract EURFiatCollateral is Collateral {
...
-            priceable = false;
+            } catch {}
+        } catch {}
+
+        // solhint-enable no-empty-blocks
+
+        if (ok) {
+            whenDefault = NEVER;
+        } else {
+            Math.min(block.timestamp + delayUntilDefault, whenDefault);
```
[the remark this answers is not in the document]
tbrent: I had this fix locally already, but I like the direction of this change. It fights against some of the chaos of the inheritance structure. Will be making it

contracts/p1/mixins/TradingLib.sol
tbrent: (proceeding assuming you mean `prepareTradeRecapitalize`)

And actually, `prepareTradeToCoverDeficit` requires the sellPrice to also be nonzero. I think this becomes an assert right after the `nextTradePair` call in `prepareTradeRecapitalize`.

tbrent: Mmm almost but not quite. Will need different constraints for the `prepareTradeSell` and `prepareTradeToCoverDeficit` cases.

PR #384, opened 2022-10-24
Fix USDT approvals - using `safeApprove`

### LINKED ISSUE #383: Make `grantRTokenAllowance` work for multiple calls to a token like USDT
USDT has a weird requirement that either the allowance being set must be 0 or that the already-set allowance is 0. 

We use `.safeIncreaseAllowance`, which won't work with this logic. We probably need to use `safeApprove`, and make multiple calls?

This also probably impacts allowances elsewhere. Should scan through rest of code and make sure everywhere we grant allowances we use a robust approach.

julianmrodri:
* Uses `safeApprove` instead of `safeIncreaseAllowance` to support the USDT case.
* Adapts both P1 and P0
*Adds a new integration test for regression testing

Resolves #383

### DIFF
--- a/contracts/p0/BackingManager.sol
+++ b/contracts/p0/BackingManager.sol
@@ -44,9 +44,8 @@ contract BackingManagerP0 is TradingP0, IBackingManager {
     /// @custom:interaction
     function grantRTokenAllowance(IERC20 erc20) external notPausedOrFrozen {
         require(main.assetRegistry().isRegistered(erc20), "erc20 unregistered");
-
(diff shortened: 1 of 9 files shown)


tbrent:
Would be useful to add a regression test to the FacadeWrite tests that include USDT in both the Prime Basket and Backup Config. Otherwise implementation LGTM

julianmrodri:
Added the regression test. It cannot go in `FacadeWrite` though, because it is more an integration test where you need to use the real USDT contract. Added a new integration test that will always run in CI to confirm it works. It was failing before the change.

julianmrodri:
Merging based on offline positive feedback.

PR #432, opened 2022-10-26
Sanitize owner input in deployer

### LINKED ISSUE #373: Sanitize owner input in Deployer
Change Deployer so that, in deploy, we require-fail if `owner = address(this)`

(Should resolve HAL-01)

julianmrodri:
* Adds checks for `owner` in DeployerP0 and DeployerP1. Includes tests.

Resolves #373


tbrent:
LGTM

PR #433, opened 2022-10-26
Prevent `maxTradeSlippage` of 100%

### LINKED ISSUE #377: Forbid maxTradeSlippage of 100%
That's a terribly silly value, and also it is a very bad case for our recapitalization algorithm.

julianmrodri:
* Allows only values < 100%
* Adapt tests

Resolves #377


tbrent:
LGTM

PR #440, opened 2022-10-28
fix endIdForVest calculation

### LINKED ISSUE #375: Incorrect edge-case handling in Facade.endIdForVest()
From the Solidified audit, item 17, the `<` in the following line should be a `<=`.

https://github.com/reserve-protocol/protocol/blob/946d9b101dd77275c6cbfe0bfe9457927bd221a9/contracts/facade/Facade.sol#L259

julianmrodri:
* Replaces `<` for `<=` in the calculations for `endIdForVest` in `FacadeRead`
* Refactors one of the tests that was used as regression testing. Confirmed the issue was occurring and added additional checks in the test to make it more robust.

Closes #375


tbrent:
LGTM

PR #447, opened 2022-11-01
FacadeAct tests

### LINKED ISSUE #441: Complete `FacadeAct` coverage tests
* The new version of `FacadeAct` has an uncovered line and a few branches that need a formal test.

julianmrodri:
* Adds new tests for branch coverage in `FacadeAct.sol`

* Adds checks in `FacadeAct.sol` to only do trading from BackingManager if the basket is not disabled. This avoids unexpected reverts when `manageTokens` is called.

             
Closes #441


contracts/facade/FacadeAct.sol
```diff
@@ -79,116 +79,128 @@ contract FacadeAct is IFacadeAct {
...
-                        return (
-                            address(cache.rTokenTrader),
-                            abi.encodeWithSelector(
-                                cache.rTokenTrader.settleTrade.selector,
-                                erc20s[i]
-                            )
-                        );
-                    }
+            } else if (cache.bh.status() != CollateralStatus.DISABLED) {
+                // Only trade if basket is sound
```
tbrent, line 83: Nit: comment should be "Only trade if basket is non-DISABLED"

tbrent: We expect there to be times where 2 tokens default slightly offset, such that the first `refreshBasket` results in IFFY

julianmrodri: Totally, will fix it

contracts/facade/FacadeAct.sol
```diff
@@ -79,116 +79,128 @@ contract FacadeAct is IFacadeAct {
...
+                    for (uint256 i = 0; i < erc20s.length; i++) {
+                        // rTokenTrader: if there's a trade to settle
+                        ITrade trade = cache.rTokenTrader.trades(erc20s[i]);
+                        if (address(trade) != address(0) && trade.canSettle()) {
+                            // cache.rTokenTrader.settleTrade(...)
                             return (
                                 address(cache.rTokenTrader),
                                 abi.encodeWithSelector(
-                                    cache.rTokenTrader.manageToken.selector,
+                                    cache.rTokenTrader.settleTrade.selector,
```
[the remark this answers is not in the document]
julianmrodri: Refactored this, only added the check for those sections that require `manageTokens` (which are the recollateralization and the sweeping revenue from Backing Manager) The rest remains the same.

contracts/facade/FacadeAct.sol
```diff
@@ -79,7 +79,9 @@ contract FacadeAct is IFacadeAct {
                         );
                     }
                 }
-            } else if (!cache.bh.fullyCollateralized()) {
+            } else if (
```
tbrent, line 82: Actually this condition should be status == CollateralStatus.SOUND. Sorry for not catching that on the first pass through. BackingManager.manageTokens reverts when IFFY

julianmrodri: So the check should be `if (cache.bh.status() == CollateralStatus.SOUND && !cache.bh.fullyCollateralized()) {` Is it the same for the call below? in Line 155?

PR #452, opened 2022-11-03
Bump yarn version

### LINKED ISSUE #435: Review `node` and `yarn` setup
We should review and adjust our required versions for `node` and `yarn`. Some users complained they had to downgrade to node 16.0 and our yarn version is not a released version.  

Maybe bumping `hardhat` and `yarn` will give us some benefits. Once this is done we should regenerate te lockfile, do a fresh install locally and push changes. Of course we need to check that everything is running as expected in regards to the dev environment.

julianmrodri:
* Bump yarn version in current setup.

Resolves #435


julianmrodri:
Merging based on offline discussion. Simply bumps yarn.

PR #497, opened 2022-11-29
Multiple beneficiaries

### LINKED ISSUE #489: Support multiple beneficiaries

julianmrodri:
* Adds support for multiple beneficiaries in `FacadeWrite`, each with its own Revenue share distribution.
* Adapts all tests and scripts.

Closes #489

### DIFF
--- a/contracts/facade/FacadeWrite.sol
+++ b/contracts/facade/FacadeWrite.sol
@@ -36,9 +36,9 @@ contract FacadeWrite is IFacadeWrite {
         // Validate beneficiaries
         for (uint256 i = 0; i < setup.beneficiaries.length; ++i) {
             require(
-                setup.beneficiaries[i].beneficiary != address(0) ||
-                    (setup.beneficiaries[i].revShare.rTokenDist == 0 &&
-                        setup.beneficiaries[i].revShare.rsrDist == 0),
+                setup.beneficiaries[i].beneficiary != address(0) &&
+                    (setup.beneficiaries[i].revShare.rTokenDist > 0 ||
+                        setup.beneficiaries[i].revShare.rsrDist > 0),
                 "beneficiary revShare mismatch"
             );
         }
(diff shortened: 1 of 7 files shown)


contracts/facade/FacadeWrite.sol
```diff
@@ -33,12 +33,15 @@ contract FacadeWrite is IFacadeWrite {
...
-            setup.beneficiary != address(0) ||
-                (setup.revShare.rTokenDist == 0 && setup.revShare.rsrDist == 0),
-            "beneficiary revShare mismatch"
-        );
+        // Validate beneficiaries
+        for (uint256 i = 0; i < setup.beneficiaries.length; ++i) {
+            require(
+                setup.beneficiaries[i].beneficiary != address(0) ||
+                    (setup.beneficiaries[i].revShare.rTokenDist == 0 &&
+                        setup.beneficiaries[i].revShare.rsrDist == 0),
```
tbrent, line 41: These requires that were here before were because you _had_ to supply a beneficiary. In the world where there are multiple beneficiaries and the array can just be zero length, then I think the require should be:
`require(setup.beneficiaries[i].beneficiary != address(0) && (setup.beneficiaries[i].revShare.rTokenDist > 0 || setup.beneficiaries[i].revShare.rsrDist > 0)`

test/FacadeWrite.test.ts
```diff
@@ -146,10 +154,11 @@ describe('FacadeWrite contract', () => {
     })
     facadeWrite = <FacadeWrite>await FacadeFactory.deploy(deployer.address)
 
-    const revShare = { rTokenDist: bn('2'), rsrDist: bn('3') } // 0.1%
+    revShare = { rTokenDist: bn('2'), rsrDist: bn('3') } // 0.5% for each beneficiary
```
tbrent, line 157: Here I'd make sure to use different share amounts for the two beneficiaries

contracts/facade/FacadeWrite.sol
julianmrodri: ok, makes sense. there is no need for a 0 distribution in the original setup

test/FacadeWrite.test.ts
julianmrodri: Ok will do

contracts/facade/FacadeWrite.sol
julianmrodri: Done

tbrent:
LGTM!

PR #513, opened 2022-12-05
RTokenAsset returning incorrect price

gjaldon:
# Bug Report

### Description 

There seems to be an issue with how RTokenAsset computes its price. If an RToken is backed by only one collateral, that RToken's price should be equal to that single collateral's price since it makes up 100% of the basket. I tried using only 1 non-fiat collateral to back an RToken and I would get an entirely different price for the RToken compared to the non-fiat collateral's price.

### Reproduction

I added a [test](https://github.com/reserve-protocol/protocol/compare/master...gjaldon:incorrect-basket-price?expand=1#diff-fb5f507b68f21557d56f36ec90c2e0f4e65c2d0b3b24b260a4a7d31702faee79R1624-R1640) in this PR that would reproduce the issue.

### DIFF
--- a/test/integration/AssetPlugins.test.ts
+++ b/test/integration/AssetPlugins.test.ts
@@ -1602,9 +1602,7 @@ describeFork(`Asset Plugins - Integration - Mainnet Forking P${IMPLEMENTATION}`,
       expect(backing[0]).to.equal(dai.address)
       expect(backing[1]).to.equal(stataDai.address)
       expect(backing[2]).to.equal(cDai.address)
-
       expect(backing.length).to.equal(3)
-
       // Check other values
       expect(await basketHandler.nonce()).to.be.gt(bn(0))
       expect(await basketHandler.timestamp()).to.be.gt(bn(0))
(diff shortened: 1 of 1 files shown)


porco-rosso-j:
Could you show us the returned values you got for the examination? I think RToken price that both RToken and RToken Asset return via `StrictPrice()` is the price of `target`.

Like, if you issue 1k RToken that has target as ETH, hence the price of one unit of this RToken is supposed to be the same as ETH, one Rtoken should be backed by apx 50k cETH. 

This is just my speculation but in the case where you configured cETH as `ref` and ETH  as`target`, this could happen even if the collateral makes up 100% of the basket. 

*I'm not from Reserve team but a participant in the hackathon.

tbrent:
> Could you show us the returned values you got for the examination? I think RToken price that both RToken and RToken Asset return via `StrictPrice()` is the price of `target`.
> 
> Like, if you issue 1k RToken that has ETH as target, hence the price of one unit of this RToken is supposed to be the same as ETH, 1k Rtoken should be backed by apx 50k cETH, not 1k cETH
> 
> This is just my speculation but in the case where you configured cETH as `ref` and ETH as`target`, this could happen even if the collateral makes up 100% of the basket.
> 
> *I'm not from Reserve team but a participant in the hackathon.

This is right, I think. cTokens are confusing because they're at 1/50th the scale of normal tokens.

gjaldon:
Thanks for the explanations @tbrent and @porco-rosso-j 🙏🏼

PR #546, opened 2022-12-16
In P0 RToken, do not refresh assets upon cancel().

fiddlemath:
This is unecessary in any case, and keeps P0 and P1 refreshing
at the same time.


julianmrodri (approved):
LGTM!

PR #547, opened 2022-12-16
pricing refactor unit test coverage

pmckelvy1:


tbrent:
@pmckelvy1 great job writing up the explanation for overflow pt2, that's precisely it

PR #555, opened 2022-12-20
add comment of untested line to FiatCollateral.

pmckelvy1:


julianmrodri (changes requested):
Please remove the  `untestable`  comment in line 186 now that you added this one.

julianmrodri:
Also I think we were using `untestable` instead of `untested`?

julianmrodri (approved):

PR #567, opened 2023-01-05
Add gas commands

akshatmittal:


julianmrodri (approved):
LGTM. Commands run successfully for me locally and solves the suggestion by C4rena.

PR #568, opened 2023-01-06
collateralShortfall: nit to make p0 and p1 revert in all the same cases

tbrent:
I assume we will not include this in the C4 hackathon


julianmrodri (approved):
LGTM

Issue #582, opened 2023-01-27
Fix initial battery block


tmattimore:

tmattimore:
must fix if possible to overflow, but it can't overflow.

Issue #581, opened 2023-01-27
Permit & Delegation should use independent nonces


tmattimore:

tmattimore:
could work, but will make ti easier for people to interact with protocol

Issue #580, opened 2023-01-27
Demote modifier for certain functions from `notPausedOrFrozen` to `notFrozen`


tmattimore:

tmattimore:
We DO want redemption to be allowed when `paused`. Not when `frozen`. 

Open question: _should_ `payoutRewards` be callable in frozen or pause?

tmattimore:
Should fix to make match docs or come up with reasoning around why docs should change.

PR #584, opened 2023-01-28
Fix Known Issues

akshatmittal:
Working PR for known issues. Solves the following:

Wave 1 Issues
- [x] #613
- [x] #599
- [x] RecollateralizationLibP1: range.bottom should be FLOOR’ed in L201
- [x] #610
- [x] Distributor: can register Furnace/StRSR a second time by using their addresses directly

Wave 2 Issues
- [x] #581
- [x] #594 (test case only)
- [x] #580 
- [x] #612 
- [x] #600
- [x] #608
- [x] #601
- [x] #582


contracts/p1/StRSR.sol
```diff
@@ -416,6 +418,8 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
         // Transfer RSR to caller
         emit ExchangeRateSet(initRate, exchangeRate());
         IERC20Upgradeable(address(rsr)).safeTransfer(_msgSender(), seizedRSR);
+
+        rsrRewardsAtLastPayout = rsrRewards();
```
tbrent, line 422: This breaks checks-effects-interactions...not sure what to do about it though. 

`rsrRewardsAtLastPayout = rsrRewards() - seizedRSR;` doesn't seem _great_

contracts/p0/BasketHandler.sol
```diff
@@ -170,7 +170,7 @@ contract BasketHandlerP0 is ComponentP0, IBasketHandler {
 
         require(
             main.hasRole(OWNER, _msgSender()) ||
-                (status() == CollateralStatus.DISABLED && !main.pausedOrFrozen()),
+                (status() == CollateralStatus.DISABLED && !main.frozen()),
```
tbrent, line 173: I wonder if we should keep this as `pausedOrFrozen()`...if the system is paused why should we support changing the basket? Recall that at least while it is DISABLED people can get the balance of DISABLED collateral token out, while if we change the basket then they won't get anything, and we won't have any of the token replacing it yet until we become unpaused and trade.

tbrent:
Did we decide not to make a change for [this](https://github.com/code-423n4/2023-01-reserve-findings/issues/397) one?

akshatmittal:
> Did we decide not to make a change for [this](https://github.com/code-423n4/2023-01-reserve-findings/issues/397) one?

No change is necessary here since OZ's ECDSA library reverts if the recovered signer is `address(0)`, I believe we decided to dispute the finding since we are using the library instead of directly using `ecrecover` but add a test case anyway.

contracts/p1/StRSR.sol
akshatmittal: Yeah, that was the previous solution, but ended up with this since it most closely follows separation of concerns while mildly breaking checks-effects-interactions.

PR #585, opened 2023-01-30
Narrow bu band

### LINKED ISSUE #590: #235 recollatarization lib causes unnecessary losses to RToken holders

### LINKED ISSUE #604: Shortfall might be calculated incorrectly if a price value for one collateral isn't fetched correctly #200

### LINKED ISSUE #609: recollateralizationLib #106

tbrent:
Closes #590 
Closes #604
Closes #609 

C4 Issues
- https://github.com/code-423n4/2023-01-reserve-findings/issues/235
- https://github.com/code-423n4/2023-01-reserve-findings/issues/106
- https://github.com/code-423n4/2023-01-reserve-findings/issues/200

### DIFF
--- a/contracts/p0/mixins/TradingLib.sol
+++ b/contracts/p0/mixins/TradingLib.sol
@@ -338,7 +338,7 @@ library TradingLibP0 {
                 // (2) Lose minTradeVolume to dust (why: auctions can return tokens)
                 // Q: Why is this precisely where we should take out minTradeVolume?
                 // A: Our use of isEnoughToSell always uses the low price (lotLow, technically),
-                //   so min trade volumes are always assested based on low prices. At this point
+                //   so min trade volumes are always assesed based on low prices. At this point
                 //   in the calculation we have already calculated the UoA amount corresponding to
                 //   the excess token balance based on its low price, so we are already set up
                 //   to straightforwardly deduct the minTradeVolume before trying to buy BUs.
--- a/contracts/p0/mixins/TradingLib.sol
+++ b/contracts/p0/mixins/TradingLib.sol
@@ -309,19 +309,19 @@ library TradingLibP0 {
             // range.bottom: contribution from balance beyond basketsHeld
             {
                 // sell at low and buy at high
-                // also assume we take maxTradeSlippage loss
-                // {BU} = {UoA/tok} * {tok} / {UoA/BU} * {1}
-                uint192 b = low.mul(bal - inBasket, FLOOR).div(basketPriceHigh).mul(
-                    FIX_ONE.minus(ctx.maxTradeSlippage),
-                    FLOOR
-                );
-
-                // {BU} = {UoA} / {UoA/BU}
-                uint192 dust = ctx.minTradeVolume.div(basketPriceLow, CEIL);
+                // {UoA} = {UoA/tok} * {tok}
+                uint192 b = low.mul(bal - inBasket, FLOOR);
 
                 // Account for potential dust loss
-                b = (b < dust) ? 0 : b - dust;
-                range.bottom += b;
+                b = (b < ctx.minTradeVolume) ? 0 : b - ctx.minTradeVolume;
+
+                // Then assume we take maxTradeSlippage loss
+                // {BU} = {UoA} * {1} / {UoA/BU}
+                range.bottom += b.mulDiv(
+                    FIX_ONE.minus(ctx.maxTradeSlippage),
+                    basketPriceHigh,
+                    FLOOR
+                );
             }
         }
 
--- a/contracts/p0/mixins/TradingLib.sol
+++ b/contracts/p0/mixins/TradingLib.sol
@@ -286,8 +286,10 @@ library TradingLibP0 {
                 (uint192 lotLow, ) = asset.lotPrice(); // {UoA/tok}
 
                 // Intentionally include value of IFFY/DISABLED collateral
-                if (inBasket == 0 && !isEnoughToSell(asset, bal, lotLow, ctx.minTradeVolume))
-                    continue;
+                if (
+                    ctx.bh.quantity(erc20s[i]) == 0 &&
+                    !isEnoughToSell(asset, bal, lotLow, ctx.minTradeVolume)
+                ) continue;
             }
 
             (uint192 low, uint192 high) = asset.price(); // {UoA/tok}
--- a/contracts/p0/mixins/TradingLib.sol
+++ b/contracts/p0/mixins/TradingLib.sol
@@ -226,29 +226,39 @@ library TradingLibP0 {
         uint192 bottom; // {BU}
     }
 
-    // It's a precondition for all the below helpers that their `erc20s` argument contains at
-    // least all basket collateral, plus any registered assets for which the BackingManager has a
-    // nonzero balance.
-
-    // This function returns a "plausible range of BUs" assuming that the trading process follows
-    //     the following rules:
+    // Compute the basket range
+    // Algorithm intuition: Trade conservatively. Quantify uncertainty based on the proportion of
+    // token balances requiring trading vs not requiring trading. Decrease uncertainty the largest
+    // amount possible with each trade.
     //
+    // How do we know this algorithm converges?
+    // Assumption: constant prices
+    // Any volume traded narrows the BU band. Why:
+    //   - We might increase `basketsHeld` from run-to-run, but will never decrease it
+    //   - We might decrease the UoA amount of excess balances beyond `basketsHeld` from
+    //       run-to-run, but will never increase it
+    //   - Balances within `basketsHeld` contribute to `range.top` and `range.bottom` equally
+    //   - Balances beyond `basketsHeld` float `range.top` and depress `range.bottom`
+    //
+    // Preconditions:
+    // - ctx is correctly populated with current basketsHeld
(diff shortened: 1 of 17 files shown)


contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -114,172 +106,120 @@ library RecollateralizationLibP1 {
...
-
-            // Accumulate potential losses to dust
-            potentialDustLoss = potentialDustLoss.plus(rules.minTradeVolume);
+            assert(high != FIX_MAX || inBasket == 0); // collateral in the basket must be priced
+
+            // throughout this section +/- is same as Fix.plus/Fix.minus
+
+            // range.top: contribution from balance beyond basketsHeld
+            {
+                // sell at high and buy at low
```
pmckelvy1, line 185: this phrasing keeps confusing me. high what? low what?

pmckelvy1 (changes requested):

pmckelvy1:
ok i have made a first pass at the code and I believe I understand the algorithm pretty well now. no glaring issues, but I'll take another look with fresh eyes first thing tomorrow and also review the tests.

contracts/p1/mixins/RecollateralizationLib.sol
pmckelvy1: `sell sellTok at its high price, and buy BU's at their low price`

test/Recollateralization.test.ts
```diff
@@ -3527,14 +3573,73 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
...
+            name: 'TradeStarted',
+            args: [anyValue, token1.address, backupToken1.address, anyValue, anyValue],
+            emitted: true,
+          },
+        ])
+
+        auctionTimestamp = await getLatestBlockTimestamp()
+
+        // Check new auction
+        // Token1 -> Backup Token 1 Auction
```
pmckelvy1, line 3595: can we guarantee that we will be able to finalize trading? what is causing 2 of the same auction to run back-to-back (token 1 => backup 1), and how do we know we won't keep running these ad infinitum?

pmckelvy1 (changes requested):

pmckelvy1:
in `basketRange`, we price the range by considering 2 trades for each side of the range:
1) sell tok for UoA
2) sell UoA for BU

but BU is (potentially) inclusive of other tokens that already fill the max of amount of BU's we need. i know that this pr is already sectioning off those assets in the first part of the function by pr-adding `basketsHeld` to the range bounds, but is there more optimization to be had by by actually dissecting the BU from these later price calculations?  i guess it then turns into an O(N) problem to look at trading for each needed token individually and leaving out the surpluses. and maybe this already gets accounted for? idk, just writing down thoughts...

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -114,172 +106,120 @@ library RecollateralizationLibP1 {
...
+            uint192 inBasket = ctx.bh.quantity(reg.erc20s[i]).mul(ctx.basketsHeld, FLOOR);
+            if (bal < inBasket) inBasket = bal; // not sure if needed
+
+            // Skip over dust-balance assets not in the basket
+            {
+                (uint192 lotLow, ) = reg.assets[i].lotPrice(); // {UoA/tok}
+
+                // Intentionally include value of IFFY/DISABLED collateral
+                if (
+                    inBasket == 0 &&
```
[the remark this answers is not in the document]
tbrent: Good catch. This should be checking whether `quantity()` is zero, not `inBasket`. Checking `inBasket` does a similar thing but not quite the same. 

The reason we calculate `inBasket` though is so we can take `bal - inBasket`.

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -114,172 +106,120 @@ library RecollateralizationLibP1 {
...
+            }
+
+            // range.bottom: contribution from balance beyond basketsHeld
+            {
+                // sell at low and buy at high
+                // {UoA} = {UoA/tok} * {tok}
+                uint192 b = low.mul(bal - inBasket, FLOOR);
+
+                // Account for potential dust loss
+                b = (b < ctx.minTradeVolume) ? 0 : b - ctx.minTradeVolume;
```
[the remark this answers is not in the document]
tbrent: It's because in the worst-case any auction could fill for the entire volume minus `minTradeVolume`. It might return `minTradeVolume` worth of sell tokens to us, and since we won't sell that amount, it becomes lost value. We have to build this into our projection of `range.bottom` ahead of time. 

More formally: for each token we sell, the maximum of `minTradeVolume` and the UoA value of the amount being sold might become inaccessible as dust.

test/Recollateralization.test.ts
tbrent: Wrote up an explanation in comments...

test/Recollateralization.test.ts
```diff
@@ -4242,11 +4348,125 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
...
+          {
+            contract: backingManager,
+            name: 'TradeSettled',
+            args: [anyValue, token1.address, backupToken2.address, sellAmt5, minBuyAmt5],
+            emitted: true,
+          },
+          {
+            contract: backingManager,
+            name: 'TradeStarted',
+            args: [anyValue, backupToken1.address, backupToken2.address, anyValue, anyValue],
```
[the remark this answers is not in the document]
tbrent: At some point the deficit of buy tokens relative to `range.bottom` is less than `minTradeVolume`, at which point `prepareTradeToCoverDeficit` rounds up the size of the auction to minTradeVolume, resulting in an overbuy and usually ending the purchasing of that asset. Basically there's a clean-up step where for each collateral we fill in the final remaining amounts. 

This isn't exactly an answer to your question, admittedly.

contracts/p0/mixins/TradingLib.sol
```diff
@@ -308,16 +325,29 @@ library TradingLibP0 {
...
 
-                // Then assume we take maxTradeSlippage loss
+                // (1) Sell tokens at low price
+                // {UoA} = {UoA/tok} * {tok}
+                uint192 val = low.mul(bal - inBasket, FLOOR);
+
+                // (2) Lose minTradeVolume to dust (why: auctions can return tokens)
+                // Q: Why is this precisely where we should take out minTradeVolume?
+                // A: Our use of isEnoughToSell always uses the low price (lotLow, technically),
+                //   so min trade volumes are always assested based on low prices. At this point
```
pmckelvy1, line 341: `assested`?

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -158,31 +168,37 @@ library RecollateralizationLibP1 {
...
 
-            assert(high != FIX_MAX || inBasket == 0); // collateral in the basket must be priced
+            assert(high != FIX_MAX || q == 0); // collateral in the basket must be priced
 
             // throughout this section +/- is same as Fix.plus/Fix.minus
 
+            // Ignore dust amounts for assets not in the basket; their value is inaccessible
+            // {tok} = {tok/BU} * {BU}
+            uint192 inBasket = q.mul(ctx.basketsHeld, FLOOR);
+            if (bal < inBasket) inBasket = bal; // not sure if needed, might be unreachable
```
pmckelvy1, line 193: i feel like if we execute this if result then something has gone horribly wrong, no? can we really ever have a balance less than whats in the current basket?

pmckelvy1 (changes requested):

contracts/p1/mixins/RecollateralizationLib.sol
tbrent: Agreed, shouldn't happen. 

`BasketHandler.basketsHeldBy` computes `basketsHeld` defensively: it does a CEIL in computation of the quantity: `uint192 q = basket.refAmts[basket.erc20s[i]].div(refPerTok, CEIL);` and a FLOOR in computation of baskets: `baskets = fixMin(baskets, bal.div(q));`(div without specification performs a FLOOR).

So if we take this `basketsHeld` and multiply by q, and take the FLOOR, this should be <= the token balance. Ok, agreed. I'm going to delete this in P1 and in P0 put an assert, and maybe fuzz finds a way to find it.

pmckelvy1: yeah i think an assert is the way to go here

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -190,7 +190,6 @@ library RecollateralizationLibP1 {
             // Ignore dust amounts for assets not in the basket; their value is inaccessible
             // {tok} = {tok/BU} * {BU}
             uint192 inBasket = q.mul(ctx.basketsHeld, FLOOR);
-            if (bal < inBasket) inBasket = bal; // not sure if needed, might be unreachable
```
pmckelvy1: no assert here as well? (like the one in p0 tradinglib?

tbrent: There will still be a revert down below when we do `bal - inBasket`. The assert in P0 is there to search for in our differential testing.

PR #617, opened 2023-01-31
don't affect strsr stakeRate until at least 1 full RSR is staked

pmckelvy1:


test/ZZStRSR.test.ts
```diff
@@ -1706,6 +1707,43 @@ describe(`StRSRP${IMPLEMENTATION} contract`, () => {
...
+      const attackerBalBefore = await rsr.balanceOf(addr1.address)
+      const stakerBalBefore = await rsr.balanceOf(addr2.address)
+
+      // stake 1 wei
+      await rsr.connect(addr1).approve(stRSR.address, bn(1))
+      await stRSR.connect(addr1).stake(bn(1))
+
+      // wait, then payout rewards
+      await advanceTime(12) // 12 second reward period
+      await stRSR.payoutRewards()
```
tbrent, line 1723: I think before and after this line we want to read out `exchangeRate()` and assert it does not change

tbrent: Also we've had trouble with `advanceTime` when dealing with small numbers. Might want to use `await setNextBlockTimestamp((await getLatestBlockTimestamp()) + 12)`

pmckelvy1:
fixes #583

PR #619, opened 2023-02-01
fix rtoken exchange rate melt bug

pmckelvy1:


pmckelvy1:
fixes #616

test/RToken.test.ts
```diff
@@ -1338,13 +1344,13 @@ describe(`RTokenP${IMPLEMENTATION} contract`, () => {
...
     })
 
-    it('Should not allow melt to set BU exchange rate to below 1e-9', async () => {
-      // melt()
-      await expect(
-        rToken.connect(addr1).melt(issueAmount.sub(issueAmount.div(bn('1e9'))).add(1))
-      ).to.be.revertedWith('BU rate out of range')
-      await rToken.connect(addr1).melt(issueAmount.sub(issueAmount.div(bn('1e9'))))
-    })
+    // it('Should not allow melt to set BU exchange rate to below 1e-9', async () => {
```
tbrent, line 1347: We should keep this test, just need to do a larger initial mint for it

contracts/p1/RToken.sol
```diff
@@ -274,6 +274,7 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
     //   totalSupply' = totalSupply - amtRToken
     function melt(uint256 amtRToken) external notPausedOrFrozen {
         _burn(_msgSender(), amtRToken);
+        require(totalSupply() >= FIX_ONE, "rToken supply too low to melt");
```
tbrent, line 277: I was imagining the require being before the burn rather than after. Do we know a reason for it to be one way vs another?

pmckelvy1:
fixes https://github.com/code-423n4/2023-01-reserve-findings/issues/384

test/RToken.test.ts
pmckelvy1: i don't think its possible to reach that error when melting via the new code (the new require catches first)

tbrent: The new require throws when the remaining balance after melting is < FIX_ONE. We can still get to the 1e9 and 1e-9 exchange rates, we just need to start with a much higher totalSupply. I think if supply is 1e28 we can set basketsNeeded to 1e19-1 and 1e37+1.

PR #620, opened 2023-02-02
Universal revenue hiding

### LINKED ISSUE #587: Introduce universal revenue hiding for base collateral plugin
Tradeoffs
- RToken inherits price uncertainty (aka peg off by 1/1,000,000 if that is number we choose).
- No matter what number we choose, it'll never be enough because each plugin doesn't know. 
- If there are defaults as a result of this edge case, then there will be false positives, which will lead to slightly extra RSR sales.

Reasons to do on all plugins vs just Compound:
- We have been looking at these forever and we just realized Compound... other ones are likely as well?
- If deploy contract to mainnet and watch the contract for defaults... only works for things like Compound, not like uniswap.

Chosen Approach:
- Implement at base collateral plugin contract.
- Each plugin can opt out.

tbrent:
Closes #587 

C4
- https://github.com/code-423n4/2023-01-reserve-findings/issues/310


contracts/plugins/assets/NonFiatCollateral.sol
```diff
@@ -43,7 +43,7 @@ contract NonFiatCollateral is FiatCollateral {
         uint192 pricePerTarget = uoaPerTargetFeed.price(oracleTimeout); // {UoA/target}
 
         // {UoA/tok} = {UoA/target} * {target/ref} * {ref/tok}
```
akshatmittal, line 45: Should clarify comment that this is for when {ref/tok} is 1 which is why it's omitted.

Ideally, we should include refPerTok here to account for plugins that inherit this contract and implement their own refPerTok.

PR #622, opened 2023-02-03
melt/payoutRewards before updating ratio

### LINKED ISSUE #595: Fix inconsistent melting behavior by melting before any parameter change

pmckelvy1:
fixes #595 
fixes https://github.com/code-423n4/2023-01-reserve-findings/issues/380
fixes https://github.com/code-423n4/2023-01-reserve-findings/issues/287

### DIFF
--- a/contracts/p0/Furnace.sol
+++ b/contracts/p0/Furnace.sol
@@ -50,7 +50,8 @@ contract FurnaceP0 is ComponentP0, IFurnace {
     /// Ratio setting
     /// @custom:governance
     function setRatio(uint192 ratio_) public governance {
-        melt();
+        // solhint-disable-next-line no-empty-blocks
+        try this.melt() {} catch {}
         require(ratio_ <= MAX_RATIO, "invalid ratio");
         // The ratio can safely be set to 0, though it is not recommended
         emit RatioSet(ratio, ratio_);
--- a/contracts/p0/Furnace.sol
+++ b/contracts/p0/Furnace.sol
@@ -30,7 +30,7 @@ contract FurnaceP0 is ComponentP0, IFurnace {
 
     /// Performs any melting that has vested since last call.
     /// @custom:refresher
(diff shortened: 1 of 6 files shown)


contracts/p0/Furnace.sol
```diff
@@ -50,6 +50,7 @@ contract FurnaceP0 is ComponentP0, IFurnace {
     /// Ratio setting
     /// @custom:governance
     function setRatio(uint192 ratio_) public governance {
+        melt();
```
tbrent, line 53: Governance may want to change the ratio while paused or frozen.  Currently the `melt()` call will revert. I think we want a try-catch here.

pmckelvy1: good catch

contracts/p0/StRSR.sol
```diff
@@ -524,6 +524,7 @@ contract StRSRP0 is IStRSR, ComponentP0, EIP712Upgradeable {
     }
 
     function setRewardRatio(uint192 val) public governance {
+        _payoutRewards();
```
tbrent, line 527: Here we want to do a similar thing and only payout rewards if not frozen

tbrent: Possibly cleanest is making `payoutRewards()` public and calling it from here.

tbrent: I can handle this, though, once the tests pass

pmckelvy1: the internal `_payoutRewards()` does not care if the system is paused or frozen, only the external call has the modifier

tbrent: Right but we should not hand out any rewards while frozen, even if governance changes the ratio

tbrent: Sorry, my previous suggestion was bad. We should just do `if (!main.frozen()) _payoutRewards()` and nothing more, I think

PR #628, opened 2023-02-07
Protect against dangerous downcasts

### LINKED ISSUE #627: Dangerous downcasting
At least `RToken.issue` needs a fix. The problem might be present elsewhere in the code too.

tbrent:
Fixes #627

C4
- https://github.com/code-423n4/2023-01-reserve-findings/issues/320


tbrent:
In addition to combing the codebase for `uint192(` and replacing all production p1 usages with FixLib functions, this PR also moves the `requireBUValidExchangeRate()` call that was at the end of each RToken function to a modifier, and makes it truly universal to RToken `@interaction`s. It previously was not on `issueTo` or `redeemTo`.

PR #631, opened 2023-02-08
Full test coverage

julianmrodri:
* Improve test coverage
* Minor fixes in FacadeRead and FacadeAct
* Updates CHANGELOG and snapshots


contracts/p1/RToken.sol
```diff
@@ -113,6 +113,9 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
     /// @param recipient The address to receive the issued RTokens
     /// @param amount {qRTok} The quantity of RToken to issue
     /// @custom:interaction
+    // untestable:
+    //      `else` branch of `exchangeRateIsValidAfter` (ie. revert)
+    //       It is not possible to change the BU exchange rate through issue
```
tbrent, line 118: Nit: It is possible to increase the BU exchange rate through issue, but _only_ when the exchange rate is less than FIX_ONE.

contracts/p1/RToken.sol
```diff
@@ -199,6 +202,9 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
     //     let tokenAmt = (amount * basketsNeeded / totalSupply) baskets of support for token
     //     let prorataAmt = (amount / totalSupply) * token.balanceOf(backingManager)
     //     do token.transferFrom(backingManager, caller, min(tokenAmt, prorataAmt))
+    // untestable:
+    //      `else` branch of `exchangeRateIsValidAfter` (ie. revert)
+    //       It is not possible to change the BU exchange rate through redeem
```
tbrent, line 207: same: possible to increase BU exchange rate, but only when below FIX_ONE

test/plugins/Collateral.test.ts
```diff
@@ -622,29 +625,153 @@ describe('Collateral contracts', () => {
...
+        expect(await coll.lastSave()).to.equal(currBlockTimestamp)
+
+        // Restore value in Oracle for next collateral
+        await setOraclePrice(coll.address, bn('1e8'))
+      }
+    })
+
+    it('Should not save prices if try/price returns unpriced - Appreciating Collateral', async () => {
+      const UnpricedAppreciatingFactory = await ethers.getContractFactory(
+        'UnpricedAppreciatingFiatCollateralMock'
```
tbrent, line 723: I think we could delete `UnpricedAppreciatingFiatCollateralMock` contract and just set oracle price to something out of range.

julianmrodri: Unfortunately we do need it because it is testing this line: https://github.com/reserve-protocol/protocol/blob/66b30dd20715838acc7db29a8d5df5212cdc2143/contracts/plugins/assets/AppreciatingFiatCollateral.sol#L117

There is a difference between having `try.price` fail and having a Collateral that returns [0,FIX_MAX] but returns the price without reverting. This is the case im covering which was only covered for `FiatCollateral`.  

Unless we want to remove that condition in the Plugin or leave it uncovered.

contracts/p1/RToken.sol
julianmrodri: So what do you suggest as comment? Something like:   `BU exchange can increase, but only when less than FIX_ONE.` Would be OK?

test/plugins/Collateral.test.ts
julianmrodri: Fun fact: The one that covers it for `FiatCollateral` is the `EUR Collateral` which returns [0, FIX_MAX] if the 2nd oracle returns 0.

contracts/p1/RToken.sol
tbrent: Sounds good to me. And it cannot decrease.

test/plugins/Collateral.test.ts
tbrent: Gotcha, that is different. I see.

PR #632, opened 2023-02-08
Facade Monitor

akshatmittal:
Edit: This also has CRLF normalization.


tbrent:
Looks good! Only two comments are (i) what patrick brought up, that we want the ERC20 addresses and not the trade contract addresses, and (ii) maybe it's worth an interface?

akshatmittal:
@tbrent Updated the contract to returns ERC20s now!

tbrent:
LGTM

PR #640, opened 2023-02-17
1inch zapper proof of concept

jankjr:
# Overview

There are two main contracts at work. The Zapper and the ZapperExecutor.

The Zapper pulls funds from user, and that user got the amount of tokens requested.

ZapperExecutor is designed a bit like a multicall, but it executes whatever inner loop specified submitted in `.commands` field.


pmckelvy1 (changes requested):
the base implementation looks great! I'm not seeing any security holes in the Zapper / ZapperExecutor logic.  it seems we're putting most of the heavy lifting on the typescript side, which makes sense.

I think next steps for this pr are:
1) some basic unit tests for the sol code
2) expanding & formalizing the typescript SDK (maybe this is a separate chunk of work, so i'd be fine merging this pr and making a separate set of tickets for the typescript work)x

PR #656, opened 2023-02-20
enforce redemption throttle > issuance throttle

### LINKED ISSUE #643: throttle issuance limit must be greater than redemption limit

tbrent:
Resolves #643 

C4
- https://github.com/code-423n4/2023-02-reserve-mitigation-contest-findings/issues/79

### DIFF
--- a/contracts/p0/RToken.sol
+++ b/contracts/p0/RToken.sol
@@ -59,8 +59,8 @@ contract RTokenP0 is ComponentP0, ERC20PermitUpgradeable, IRToken {
         __ERC20Permit_init(name_);
 
         mandate = mandate_;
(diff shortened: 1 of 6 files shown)


tbrent:
I think I've come around to believing this would be over-engineering. It makes dealing with updating the throttles super annoying to think about, and this seems like something governance/a frontend could get wrong (let's put aside the fact that they should simulate). 

If an RToken really wants to have an issuance throttle that exceeds the redemption throttle, let them. If griefing starts to occur, Governance can act. 

I'll leave the PR up for a bit longer so we can decide.

tbrent:
Not enforcing at smart contract layer, closing PR

PR #661, opened 2023-02-21
Exhuastive box

pmckelvy1:
created against rc4 because it has rc4 merged into it. either merge into rc4 first or change target to master once rc4 is merged.


tbrent:
The Furnace tests are skipped, no?

pmckelvy1:
> 

good catch, just ran it though and its good 
image

tbrent:
@pmckelvy1 I made some minor documentation/comment clarifications and also unset the `SLOW` var in the `yarn test:extreme` target, since it isn't used by the scripts and is probably better suited to a human running the extreme tests locally. Ready to merge once you confirm these changes make sense. 

Also it looks like my prettier autoformat made some minor newline/end-of-line changes. Make sure I'm not playing tug-of-war and undoing something your editor will just add back-in on next save. I checked and I _think_ my autoformat is using our local `.prettierrc`, which should be correct.

LGTM

PR #664, opened 2023-02-22
Oz 4.7.3 governance

tbrent:
OZ changed how `quorum` + `proposalThreshold` work

### DIFF
--- a/contracts/plugins/governance/Governance.sol
+++ b/contracts/plugins/governance/Governance.sol
@@ -56,7 +56,7 @@ contract Governance is
         return super.votingPeriod();
     }
 
-    /// @return The number of votes required in order for a voter to become a proposer
+    /// @return {qStRSR} The number of votes required in order for a voter to become a proposer
     function proposalThreshold()
         public
         view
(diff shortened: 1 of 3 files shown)


contracts/plugins/governance/Governance.sol
```diff
@@ -27,6 +27,9 @@ contract Governance is
     GovernorVotesQuorumFraction,
     GovernorTimelockControl
 {
+    // 100%
+    uint256 public constant ONE_HUDRED_PERCENT = 1e8; // {micro %}
```
julianmrodri, line 31: HUNDRED?

contracts/plugins/governance/Governance.sol
```diff
@@ -141,19 +149,13 @@ contract Governance is
         return super._executor();
     }
 
-    /// @return {micro %} The portion of the StRSR supply the account had at a previous blocknumber
+    /// @return {qStRSR} The voting weight the account had at a previous block number
```
julianmrodri, line 152: Why did we move away from micropercents? Wasnt that on purpose that we implemented it in that way?

```
* Note that due to the elastic supply of StRSR, proposalThreshold is handled
 *   very differently than the typical approach. It is in terms of micro %,
 *   as is _getVotes().
 ```

julianmrodri (approved):
LGTM

PR #681, opened 2023-02-24
Fuzz simplification

pmckelvy1:
A pr for honing in our fuzzing tests.  We are:
- removing/internalizing some functions from some scenarios, which should decrease the number of possible permutations of functions in each sequence
- using % math instead of require statements to speed up valid scenario generation
- adding scenarion-specific scripts that use special echidna-parade flags to always include certain functions (ie. it doesn't make sense to run a rebalancing scenario if we cannot start or finish a rebalance, so we should always include the functions that do those things)

There is probably some more trial and error to be done here, so I'd like to leave this pr open for the time being until i can really test the effects of some of these changes.  I'll remove the `Do not merge` label once it seems like we might have some improvements.


pmckelvy1:
closed for now to cleanup PR list. will re-open at a later date after more testing.

PR #682, opened 2023-02-25
chore: uncomment FORK=1 in .env.example

larrythecucumber321:
Collateral plugin tests require the `FORK=1` flag to be set to be work (otherwise they get skipped), so I think this should be uncommented by default so collateral plugin devs don't get confused why their tests don't run

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,7 +24,7 @@ ETHERSCAN_API_KEY=ABC123ABC123ABC123ABC123ABC123ABC1
 # SLOW=1
 
 # Run tests and scripts using Mainnet forking - required for integration tests
-# FORK=1
+FORK=1
 
 # Block to use as default for mainnet forking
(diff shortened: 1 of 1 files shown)


tbrent:
Hmmm we should figure out a different way to handle this. For a dev just starting to develop on our repo this will cause the unit tests to fail in spooky ways they aren't obvious in their cause. Not sure what's right here, will return to later.

pmckelvy1:
i would agree with taylor that having unit tests pass is probably the preferable default.  I think a better path here is to work on a simple document like `contributing-collateral-plugins.md`, that can have this FORK detail, as well as our defined path to building / testing / submitting collateral plugins (more developer-process oriented, and it can point to `collateral.md` for the nitty-gritty details of how collateral is defined in the protocol, and how the plugins should work)

pmckelvy1:
closing so we can handle in #707

PR #691, opened 2023-03-07
LSD Plugins

pmckelvy1:


contracts/interfaces/IAsset.sol
```diff
@@ -45,16 +45,27 @@ interface IAsset is IRewardable {
 
     /// @param {UoA} The max trade volume, in UoA
     function maxTradeVolume() external view returns (uint192);
-}
 
-interface TestIAsset is IAsset {
```
tbrent, line 50: What's the reason for getting rid of the TestIAsset interface? imo the `IAsset` interface should be the interface that the _protocol_ requires of an asset. This is a lot smaller than what we currently have. For example, not every asset is going to need or have:
- `chainlinkFeed()`
- `oracleTimeout()`
- `tryPrice()`

An asset that is priced using TWAPs does not have a concept of an oracleTimeout, because it always has a price.

tbrent: Is there a cost to the `TestIAsset` interface I'm not seeing?

contracts/interfaces/IAsset.sol
```diff
@@ -105,9 +116,7 @@ interface ICollateral is IAsset {
 
     /// @return {target/ref} Quantity of whole target units per whole reference unit in the peg
     function targetPerRef() external view returns (uint192);
-}
 
-interface TestICollateral is ICollateral, TestIAsset {
```
tbrent, line 110: Same thing for `ICollateral`. Not every collateral will have a `delayUntilDefault()` or `whenDefault()`..

contracts/plugins/ankr/AnkrStakedEthCollateral.sol
```diff
@@ -0,0 +1,61 @@
...
+ * UoA = USD
+ */
+contract AnkrStakedEthCollateral is AppreciatingFiatCollateral {
+    using OracleLib for AggregatorV3Interface;
+    using FixLib for uint192;
+
+    constructor(CollateralConfig memory config, uint192 revenueHiding)
+        AppreciatingFiatCollateral(config, revenueHiding)
+    {
+        exposedReferencePrice = _underlyingRefPerTok().mul(revenueShowing);
```
tbrent, line 25: this prevents extending the plugin and overriding `_underlyingRefPerTok()`

contracts/plugins/ankr/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Functions
+
+#### refPerTok {ref/tok}
+
+The exchange rate between ETH and ankrETH can be fetched using the ankrETH contract function `ratio()`. From this, we can obtain the inverse rate from ankrETH to ETH, and use that as `refPerTok`.
+
+This new ratio, increases over time, which means that the amount of ETH redeemable for each ankrETH token always increases.
+
+`ratio()` returns the exchange rate in 10\*\*18.
```
tbrent, line 29: What does this notation mean?

contracts/plugins/assets/RTokenAsset.sol
```diff
@@ -70,7 +80,7 @@ contract RTokenAsset is IAsset {
     /// @return {UoA/tok} The lower end of the price estimate
     /// @return {UoA/tok} The upper end of the price estimate
     function price() public view virtual returns (uint192, uint192) {
-        try this.tryPrice() returns (uint192 low, uint192 high) {
+        try this.tryPrice() returns (uint192 low, uint192 high, uint192) {
```
tbrent, line 83: This is a good example why `tryPrice()` doesn't need to be on the interface. It's an internal detail of the implementation of the contract.

scripts/deployment/phase2-assets/collaterals/deploy_lido_wsteth_collateral.ts
```diff
@@ -0,0 +1,84 @@
...
+
+    const LidoStakedEthCollateralFactory: ContractFactory = await hre.ethers.getContractFactory(
+      'LidoStakedEthCollateral'
+    )
+
+    const collateral = <LidoStakedEthCollateral>await LidoStakedEthCollateralFactory.connect(deployer).deploy(
+      {
+        priceTimeout: priceTimeout.toString(),
+        chainlinkFeed: networkConfig[chainId].chainlinkFeeds.ETH,
+        oracleError: fp('0.005').toString(), // 0.5%,
```
tbrent, line 52: This oracle error should be the combined error from the two feeds. Take a look at the `combinedError()` function in `scripts/deployment/utils`.

scripts/deployment/phase2-assets/collaterals/deploy_lido_wsteth_collateral.ts
```diff
@@ -0,0 +1,84 @@
...
+        erc20: networkConfig[chainId].tokens.wstETH,
+        maxTradeVolume: fp('1e3').toString(), // 1k $ETH,
```
tbrent, line 54: Surely we don't want to cap trades involving wstETH to $1k...we've used $1m everywhere else.

scripts/deployment/phase2-assets/collaterals/deploy_lido_wsteth_collateral.ts
```diff
@@ -0,0 +1,84 @@
...
+        oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr,
+        targetName: hre.ethers.utils.formatBytes32String('ETH'),
+        defaultThreshold: fp('0.15').toString(), // 15%
+        delayUntilDefault: bn('86400').toString() // 24h
+      },
+      bn('1e14'), // revenueHiding = 0.01%
+      networkConfig[chainId].chainlinkFeeds.stETH, // targetPerRefChainlinkFeed
+      oracleTimeout(chainId, '3600').toString() // targetPerRefChainlinkTimeout
```
tbrent, line 62: This is the timeout for `STETH / USD`, not `STETH / ETH`. Should be 86400.

scripts/deployment/phase2-assets/collaterals/deploy_lido_wsteth_collateral.ts
```diff
@@ -0,0 +1,84 @@
...
+        chainlinkFeed: networkConfig[chainId].chainlinkFeeds.ETH,
+        oracleError: fp('0.005').toString(), // 0.5%,
+        erc20: networkConfig[chainId].tokens.wstETH,
+        maxTradeVolume: fp('1e3').toString(), // 1k $ETH,
+        oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr,
+        targetName: hre.ethers.utils.formatBytes32String('ETH'),
+        defaultThreshold: fp('0.15').toString(), // 15%
+        delayUntilDefault: bn('86400').toString() // 24h
+      },
+      bn('1e14'), // revenueHiding = 0.01%
```
tbrent, line 60: This is a nit, granted, but I think `fp('1e-4')` is clearer.

contracts/plugins/lido/LidoStakedEthCollateral.sol
```diff
@@ -0,0 +1,71 @@
...
+        view
+        override
+        returns (
+            uint192 low,
+            uint192 high,
+            uint192 pegPrice
+        )
+    {
+        // {UoA/target}
+        uint192 pricePerTarget = chainlinkFeed.price(oracleTimeout);
```
tbrent, line 53: I see 3 chainlink official feeds for StETH. The `StETH / ETH` feed has the widest error of them all. We could consider using the `STETH / USD` feed directly in `tryPrice()`. It would reduce compounding error. Then we could use the `STETH / ETH` feed only in the refresh() function to check for the peg.

scripts/deployment/phase2-assets/collaterals/deploy_rocket_pool_reth_collateral.ts
```diff
@@ -0,0 +1,85 @@
...
+        erc20: networkConfig[chainId].tokens.rETH,
+        maxTradeVolume: fp('1e3').toString(), // 1k $ETH,
+        oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr,
+        targetName: hre.ethers.utils.formatBytes32String('ETH'),
+        defaultThreshold: fp('0.15').toString(), // 15%
+        delayUntilDefault: bn('86400').toString() // 24h
+      },
+      bn('1e14'), // revenueHiding = 0.01%
+      networkConfig[chainId].chainlinkFeeds.rETH, // targetPerRefChainlinkFeed
+      oracleTimeout(chainId, '3600').toString() // targetPerRefChainlinkTimeout
```
tbrent, line 63: The comment here should be `refPerTokChainlinkFeed` and the value should be 86400.

scripts/deployment/phase2-assets/collaterals/deploy_rocket_pool_reth_collateral.ts
```diff
@@ -0,0 +1,85 @@
...
+
+    const RethCollateralFactory: ContractFactory = await hre.ethers.getContractFactory(
+      'RethCollateral'
+    )
+
+    const collateral = <RethCollateral>await RethCollateralFactory.connect(deployer).deploy(
+      {
+        priceTimeout: priceTimeout.toString(),
+        chainlinkFeed: networkConfig[chainId].chainlinkFeeds.ETH,
+        oracleError: fp('0.005').toString(), // 0.5%,
```
tbrent, line 53: Should be compounded error, taking into account the 2% error from `RETH / ETH` feed.

scripts/deployment/phase2-assets/collaterals/deploy_rocket_pool_reth_collateral.ts
```diff
@@ -0,0 +1,85 @@
...
+        erc20: networkConfig[chainId].tokens.rETH,
+        maxTradeVolume: fp('1e3').toString(), // 1k $ETH,
```
tbrent, line 55: $1m

scripts/deployment/phase2-assets/collaterals/deploy_rocket_pool_reth_collateral.ts
```diff
@@ -0,0 +1,85 @@
...
+      oracleTimeout(chainId, '3600').toString() // targetPerRefChainlinkTimeout
+    )
+    await collateral.deployed()
+    await collateral.refresh()
+
+    console.log(
+        `Deployed Fiat Collateral to ${hre.network.name} (${chainId}): ${collateral.address}`
+    )
+
+    assetCollDeployments.collateral.wstETH = collateral.address
```
tbrent, line 72: Typo

scripts/deployment/phase2-assets/collaterals/deploy_rocket_pool_reth_collateral.ts
```diff
@@ -0,0 +1,85 @@
...
+    // Get phase1 deployment
+    const phase1File = getDeploymentFilename(chainId)
+    if (!fileExists(phase1File)) {
+        throw new Error(`${phase1File} doesn't exist yet. Run phase 1`)
+    }
+    // Check previous step completed
+    const assetCollDeploymentFilename = getAssetCollDeploymentFilename(chainId)
+    const assetCollDeployments = <IAssetCollDeployments>getDeploymentFile(assetCollDeploymentFilename)
+
+    // Get Oracle Lib address if previously deployed (can override with arbitrary address)
```
tbrent, line 40: Can delete this comment

test/plugins/individual-collateral/collateralTests.ts
```diff
@@ -142,72 +152,65 @@ export default function fn<X extends CollateralFixtureContext>(
       })
 
       describe('prices', () => {
-        it('prices change as USDC feed price changes', async () => {
-          const clData = await chainlinkFeed.latestRoundData()
-          const decimals = await chainlinkFeed.decimals()
+        itCheckPriceChanges('prices change as USD feed price changes', async () => {
```
tbrent, line 155: I like this pattern you landed on 👍

test/plugins/individual-collateral/lido/constants.ts
```diff
@@ -0,0 +1,19 @@
+import { bn, fp } from '../../../../common/numbers'
+
+// Mainnet Addresses
+export const RSR = '0x320623b8e4ff03373931769a31fc52a4e78b5d70'
```
tbrent, line 4: In general I don't see a reason to redeclare all this data. A lot of it is in `common/configuration` and we could just import from there and access the `networkConfig`

tbrent:
In addition to the comments I left, I also think we could cut down on the surface area of the vendored interfaces. Check out `contracts/plugins/assets/ICToken.sol`. The interface should be the minimal set of functions the plugin contract requires, imo.

contracts/plugins/ankr/AnkrStakedEthCollateral.sol
julianmrodri: Interesting. I havent seen these as extendable but makes sense. What would be a solution? I think we need to do it in more than one of these plugins.

test/plugins/individual-collateral/lido/constants.ts
julianmrodri: I agree. The question is do we move all to `common/configuration` or only the repeated ones for now or the token addresses only (which are required for deployment script as well)  and not the "auxiliary" ones?

julianmrodri:
Also in the `assets` folder we still have the CompoundV2 and AaveV2 plugins. while the others are in individual folder at the root `plugins` level. Should we add `compoundv2` and `aavev2` and refactor those files? In `assets` we would only have abstract ones plus some exceptions like WBTC and WETH.

contracts/interfaces/IAsset.sol
pmckelvy1: yeah i guess i didn't think of the TWAP case since we don't have any of those yet. we can revert on this if you think it makes more sense to keep TestIAsset

contracts/interfaces/IAsset.sol
pmckelvy1: which ones wouldn't? I feel like theyre standard values for soft => hard defaults?

contracts/plugins/ankr/AnkrStakedEthCollateral.sol
pmckelvy1: so should we just say "the plugin must be refreshed after it is deployed"?

scripts/deployment/phase2-assets/collaterals/deploy_lido_wsteth_collateral.ts
pmckelvy1: makes sense

tbrent:
> Also in the `assets` folder we still have the CompoundV2 and AaveV2 plugins. while the others are in individual folder at the root `plugins` level. Should we add `compoundv2` and `aavev2` and refactor those files? In `assets` we would only have abstract ones plus some exceptions like WBTC and WETH.

Good point. Time to group under their own folder I think.

scripts/deployment/phase2-assets/collaterals/deploy_lido_wsteth_collateral.ts
pmckelvy1: yeah i think when i changed this i was translating "1k eth @ $1k is $1m" which is something i've had to do in other protocols... will revert

test/plugins/individual-collateral/lido/constants.ts
tbrent: I would keep the auxiliary ones here.

contracts/interfaces/IAsset.sol
tbrent: I do think we should do that.

contracts/interfaces/IAsset.sol
tbrent: Not all tokens have a bridge component that require a soft default check. We ended up by laziness keeping around a `delayUntilDefault` for `SelfReferentialCollateral`, but it's not a necessity. When other protocol teams develop plugins they should be free to not have this parameter. 

And for `whenDefault()`, that just comes out of inheriting from our parent classes, which isn't a necessity. `status()` is the real thing; `whenDefault()` just happened to come along for the ride

contracts/plugins/lido/LidoStakedEthCollateral.sol
tbrent: Screen Shot 2023-03-17 at 11 24 13 AM

see here: https://docs.chain.link/data-feeds/price-feeds/addresses

tbrent:
All done I think. Still have to add revenue hiding tests...will see if I can get around to that later

PR #692, opened 2023-03-09
re-enable ci reset

julianmrodri:
* Enable tests that were skipped now that Hardhat Reset works on CI


tbrent:
I've added the extreme tests + plugin integration tests to the github workflow, and fixed a test that was failing silently because these were missing.

PR #693, opened 2023-03-10
Plugin Lido WSTETH

julianmrodri:
* Adds Lido plugin (StETH)
* Includes tests
* Modifies test harness to add a new parameter to skip tests

### DIFF
--- a/contracts/plugins/lido/LidoStakedEthCollateral.sol
+++ b/contracts/plugins/lido/LidoStakedEthCollateral.sol
@@ -1,0 +1,73 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+pragma solidity 0.8.17;
+
+import "@openzeppelin/contracts/utils/math/Math.sol";
+import "contracts/plugins/assets/AppreciatingFiatCollateral.sol";
+import "contracts/plugins/lido/IWSTETH.sol";
+import "contracts/plugins/assets/OracleLib.sol";
+import "contracts/libraries/Fixed.sol";
+
+/**
+ * @title Lido Staked Eth Collateral
+ * @notice Collateral plugin for Lido stETH,
+ * tok = wstETH  (wrapped stETH)
+ * ref = stETH (pegged to ETH 1:1)
+ * tar = ETH
+ * UoA = USD
+ */
+contract LidoStakedEthCollateral is AppreciatingFiatCollateral {
+    using OracleLib for AggregatorV3Interface;
+    using FixLib for uint192;
+
+    AggregatorV3Interface public immutable targetUnitChainlinkFeed; // {UoA/target}
+    uint48 public immutable targetUnitOracleTimeout; // {s}
+
+    constructor(
+        CollateralConfig memory config,
+        uint192 revenueHiding,
+        AggregatorV3Interface targetUnitChainlinkFeed_,
+        uint48 targetUnitOracleTimeout_
+    ) AppreciatingFiatCollateral(config, revenueHiding) {
+        require(address(targetUnitChainlinkFeed_) != address(0), "missing targetUnit feed");
+        require(targetUnitOracleTimeout_ > 0, "targetUnitOracleTimeout zero");
+        targetUnitChainlinkFeed = targetUnitChainlinkFeed_;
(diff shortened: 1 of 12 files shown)


contracts/plugins/lido/README.md
```diff
@@ -0,0 +1,35 @@
+# Lido stETH Collateral Plugin
+
+## Summary
+
+This plugin allows `wstETH` holders use their tokens as collateral in the Reverse Protocol.
+
+As described in the [Lido Site](https://docs.lido.fi/guides/steth-integration-guide#wsteth) , `wstETH` is a LSD (Liquid staking derivatives) which enables users to sell or transfer stacked ETH even before withdrawal being enabled.
+
+`wstETH` will accrue revenue from **staking rewards** into itself by **increasing** the exchange rate of the `wstETH` per `stETH`.
```
pmckelvy1, line 9: "...the exchange rate of `stETH` per `wstETH`"

contracts/plugins/lido/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+You can get exchange rate from [`wstETH.stEthPerToken()`](https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract#F10) method of wstETH contract.
+
+`wstETH` contract: <https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#code>
+
+`stETH` contract: <https://etherscan.io/token/0xae7ab96520de3a18e5e111b5eaab095312d7fe84#code>
+
+As described on `stETH Intro`, `stETH` is a rebasing token which means token balances is not fixed. Due to that, it can not be used directly as collateral the Reverse Protocol. The solution for that is using `wstETH` instead which is not a rebasing token and behaves like `rETH` and `cbETH`.
```
pmckelvy1, line 17: Reserve (misspelled)

contracts/plugins/lido/LidoStakedEthCollateral.sol
```diff
@@ -0,0 +1,73 @@
...
+        view
+        override
+        returns (
+            uint192 low,
+            uint192 high,
+            uint192 pegPrice
+        )
+    {
+        // Get current market peg {eth/steth}, but the intended {target/ref} will be returned
+        pegPrice = chainlinkFeed.price(oracleTimeout); // {target/ref}
```
pmckelvy1, line 53: not something not necessarily change in this pr, just dropping a thought....

i don't think we've standardized what `chainlinkFeed` means across collateral types...sometimes it is used as UoA/target, sometimes it is used as target/ref.  I think it might be worth standardizing and renaming so that we can be clear about what it represents, as it could lead to confusion down the road otherwise.

julianmrodri: Ive looked at the ones we have that use additional feeds for example the NonFiatcollateral (wBTC) and in that case the default one is the peg, and the additional is the Target feed. So I did it in that same way. Buy I can revert them if preferred. (I think we should revert them in wBTC as well if we do that probably).

contracts/plugins/lido/README.md
```diff
@@ -0,0 +1,35 @@
...
+
+| tok    | ref   | target | UoA |
+| ------ | ----- | ------ | --- |
+| wstETH | stETH | ETH    | USD |
+
+### Functions
+
+#### refPerTok {ref/tok}
+
+This function returns rate of `ETH/wstETH`, getting from [stEthPerToken()](https://etherscan.io/token/0x7f39c581f595b53c5cb19bd0b3f8da6c935e2ca0#readContract#F10) function in wstETH contract.
```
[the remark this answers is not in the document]
julianmrodri: Fixed

contracts/plugins/lido/README.md
julianmrodri: Fixed

contracts/plugins/lido/LidoStakedEthCollateral.sol
julianmrodri: We should definitely sync them all and be consistent. Lets have this discussion before getting any of these to master next week.

PR #698, opened 2023-03-13
Update mainnet addresses & add collateral plugin addresses

larrythecucumber321:
This PR updates [README.md](https://github.com/reserve-protocol/protocol/blob/feat/collateral-script/README.md#mainnet-addresses-v200) with the most recent deployment addresses.

This additionally adds a file documenting the deployed collateral plugin addresses and their configuration parameters, found [here](https://github.com/reserve-protocol/protocol/blob/feat/collateral-script/docs/plugin-addresses.md). A script for updating/generating this table is included.


scripts/collateral-params.ts
```diff
@@ -0,0 +1,89 @@
...
+  printTable(header, body)
+}
+
+main().catch((error) => {
+  console.error(error)
+  process.exitCode = 1
+})
+
+// Helper function for printing MD table
+// This could've come from a library, but I didn't want to add a dependency
```
tbrent, line 52: 👏

tbrent:
Thanks @larrythecucumber321

PR #699, opened 2023-03-13
new DAI/USDC/aUSDC/cUSDC collateral addresses

tbrent:
New addresses for DAI/USDC/aUSDC/cUSDC collateral plugins.


tbrent:
Shouldn't merge until CompoundV2 unpauses USDC deposits. Someone will need to call `refresh()` at that point.

tbrent:
They've unpaused and we've called refresh()

PR #700, opened 2023-03-14
Trade Clearance

akshatmittal:
Adjusting buy amount instead of sell since EasyAuction operates on the buy token.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -30,3 +30,7 @@ scripts/addresses/31337*
 *.orig
 
 .idea
+
+# Scripts for local/test interactions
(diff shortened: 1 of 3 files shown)


tbrent:
LGTM. I contemplated adding CEIL rounding to the `clearingPrice` expression but ultimately decided it was wrong. I can imagine a situation where the auction platform _does_ give us a lower price than the `worstCasePrice`, and then our `boughtAmt + 1` logic ends up bringing this _just_ below the worstCasePrice, and then that rounding causes us not to detect the violation. Merging!

PR #703, opened 2023-03-14
Facade read additions

tbrent:
- Add a USD measure of deposits required to issue to `FacadeRead.issue()`
- Add new `FacadeRead.traderBalances()` function


tbrent:
@lcamargof if you could confirm the new interface looks good

PR #705, opened 2023-03-15
Auction extreme tests

tbrent:


akshatmittal:
Looks good! I added some very specific tests cases for where I _expect_ that there might be issues.

LGTM!

akshatmittal (approved):

PR #719, opened 2023-03-20
Broker upgrade

julianmrodri:
* Updates Broker with two new governable params: `Gnosis` and `TradeImplementation`
* Includes tests and compatibility with P0


tbrent:
Really nice @julianmrodri, looks perfect to me, I wouldn't change a thing

PR #720, opened 2023-03-20
Add Regression Test for Gnosis Trade Plugin

akshatmittal:


tbrent:
LGTM, assuming you ran this without the `setCode` line and the test failed, first.

PR #726, opened 2023-03-22
propose facade interface for new Register `run*Auctions` buttons

tbrent:
Proposed pattern of use:
1. `Run Recollateralization Auctions` button
  (i) First: call `FacadeRead.auctionsSettleable(backingManager)`. If return array contains an entry (it will never contain more than 1), then there is a settlement action to be taken on the backingManager. 
  (ii) Second (can be done in parallel): call `FacadeAct.canRunRecollateralizationAuctions(backingManager)`. If it returns true, then `backingManager.manageTokensSortedOrder([])` can be called productively.
  (iii) If both (i) and (ii) returned values, we'll need to package these up via a multicall. See: `backingManager.multicall()`. Otherwise, can call `settleTrade(erc20)` or `manageTokensSortedOrder([])` in isolation.

2. `Run RSR Revenue Auctions` button
  (i) First: call `FacadeRead.auctionsSettleable(rsrTrader)`. Save list of ERC20s
  (ii) Second (can be done in parallel): call `FacadeAct.getRevenueAuctionERC20s(rsrTrader)`. Save list of ERC20s
  (iii) If _either_ of the two ERC20 lists are non-empty, then `FacadeAct.runRevenueAuctions` can be called. No Multicall required here. 

(and the same would apply for the `Run RToken Revenue Auctions` button)

### DIFF
--- a/contracts/facade/FacadeAct.sol
+++ b/contracts/facade/FacadeAct.sol
@@ -351,4 +351,72 @@ contract FacadeAct is IFacadeAct {
         uint192 size = price == 0 ? FIX_MAX : minTradeVolume.div(price, ROUND);
         return size > 0 ? size : 1;
     }
+
+    /// To use this, call via callStatic.
+    /// @return toStart The ERC20s that have auctions that can be started
+    /// @custom:static-call
+    function getRevenueAuctionERC20s(IRevenueTrader revenueTrader)
+        external
+        returns (IERC20[] memory toStart)
+    {
+        Registry memory reg = revenueTrader.main().assetRegistry().getRegistry();
+
+        // Forward ALL revenue
+        revenueTrader.main().backingManager().manageTokens(reg.erc20s);
+
+        // Calculate which erc20s can have auctions started
+        uint256 num;
+        IERC20[] memory unfiltered = new IERC20[](reg.erc20s.length); // will filter down later
+        for (uint256 i = 0; i < reg.erc20s.length; ++i) {
+            // Settle first if possible. Required so we can assess full available balance
+            ITrade trade = revenueTrader.trades(reg.erc20s[i]);
+            if (address(trade) != address(0) && trade.canSettle()) {
+                revenueTrader.settleTrade(reg.erc20s[i]);
+            }
+
+            uint256 tradesOpen = revenueTrader.tradesOpen();
+
+            try revenueTrader.manageToken(reg.erc20s[i]) {
+                if (revenueTrader.tradesOpen() - tradesOpen > 0) {
+                    unfiltered[num] = reg.erc20s[i];
+                    ++num;
+                }
+            } catch {}
+        }
+
+        // Filter down
+        toStart = new IERC20[](num);
+        for (uint256 i = 0; i < num; ++i) {
+            toStart[i] = unfiltered[i];
+        }
+    }
+
+    /// To use this, first call:
+    ///   - FacadeRead.auctionsSettleable(revenueTrader)
+    ///   - getRevenueAuctionERC20s(revenueTrader)
+    /// If either arrays returned are non-empty, then can call this function.
+    /// Logic:
+    ///   For each ERC20 in `toSettle`:
+    ///     - Settle any open ERC20 trades
+    ///   For each ERC20 in `toStart`:
+    ///     - Transfer any revenue for that ERC20 from the backingManager to revenueTrader
+    ///     - Call `revenueTrader.manageToken(ERC20)` to start an auction
+    function runRevenueAuctions(
+        IRevenueTrader revenueTrader,
+        IERC20[] memory toSettle,
+        IERC20[] memory toStart
+    ) external {
+        // Settle auctions
+        for (uint256 i = 0; i < toSettle.length; ++i) {
+            revenueTrader.settleTrade(toSettle[i]);
+        }
+
+        // Transfer revenue backingManager -> revenueTrader
+        revenueTrader.main().backingManager().manageTokens(toStart);
+
+        // Start auctions
+        for (uint256 i = 0; i < toStart.length; ++i) {
+            revenueTrader.manageToken(toStart[i]);
+        }
+    }
 }
--- a/contracts/facade/FacadeAct.sol
+++ b/contracts/facade/FacadeAct.sol
@@ -353,6 +353,32 @@ contract FacadeAct is IFacadeAct {
     }
 
     /// To use this, call via callStatic.
+    /// If canStart is true, proceed to runRecollateralizationAuctions
+    /// @return canStart true iff a recollateralization auction can be started
+    /// @custom:static-call
+    function canRunRecollateralizationAuctions(IBackingManager bm)
+        external
+        returns (bool canStart)
+    {
+        IERC20[] memory erc20s = bm.main().assetRegistry().erc20s();
+
+        // Settle all backingManager auctions
+        for (uint256 i = 0; i < erc20s.length; ++i) {
+            ITrade trade = bm.trades(erc20s[i]);
+            if (address(trade) != address(0) && trade.canSettle()) {
+                bm.settleTrade(erc20s[i]);
+            }
+        }
+
+        uint256 tradesOpen = bm.tradesOpen();
+        if (tradesOpen != 0) return false;
+
+        // Try to launch auctions
+        bm.manageTokensSortedOrder(new IERC20[](0));
+        return bm.tradesOpen() - tradesOpen > 0;
+    }
+
+    /// To use this, call via callStatic.
     /// @return toStart The ERC20s that have auctions that can be started
     /// @custom:static-call
     function getRevenueAuctionERC20s(IRevenueTrader revenueTrader)
--- a/contracts/facade/FacadeAct.sol
+++ b/contracts/facade/FacadeAct.sol
@@ -375,7 +375,7 @@ contract FacadeAct is IFacadeAct {
 
         // Try to launch auctions
         bm.manageTokensSortedOrder(new IERC20[](0));
-        return bm.tradesOpen() - tradesOpen > 0;
+        return bm.tradesOpen() > 0;
     }
 
     /// To use this, call via callStatic.
--- a/test/FacadeAct.test.ts
+++ b/test/FacadeAct.test.ts
@@ -22,6 +22,7 @@ import {
   CTokenMock,
   ERC20Mock,
   FacadeAct,
+  IFacadeRead,
   FiatCollateral,
   GnosisMock,
   IAssetRegistry,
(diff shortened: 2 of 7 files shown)


tbrent:
(still needs tests. will return to once we get confirmation this design makes sense)

larrythecucumber321:
Your proposed design makes sense to me, and it would be very possible to implement this on the UI. My only thoughts are on whether we can simply this flow in the UI to ONE button. For 99% of users, none of these buttons will mean anything, nor should they ever touch them. I wouldn't want to waste screen real estate for 1% of power users.

I'm wondering if this is a reasonable method to ascertain system state at a given point in time (and then make the appropriate calls you laid out):

1. Check whether basket is fully collateralized 
    - Collateralized if `basketHandler.status() == SOUND && basketHandler.fullyCollateralized()`
    - If true, then run the `auctionsSettleable` and `getRevenueAuctionERC20s` checks on each of the Trader contracts
    - With the returned array, the UI will know which Trader and ERC20 is up to bat and that can be reflected in the button text
2. If the basket is not fully collateralized
    - `Run Recollateralization Auctions` as you've detailed

Let me know if I'm off-base with this line of thinking. Conceivably, if we wanted the UI to do as little thinking as possible, we could abstract all this away into `FacadeAct.getActCalldata()` right? One could decode this on the frontend to find the function signature and params, etc.

tbrent:
This can totally be put in a single button on the frontend! Good idea. I don't want to prescribe exact UI layouts, that was probably overstepping. Just wanted to demonstrate how the two different flows could be executed.

tbrent:
The issue with the `getActCalldata` approach is that it's not able to bundle the settlement of an auction together with the starting of the next one. Or in the revenue case, it's not able to forward tokens from the BackingManager to the RevenueTrader and start the revenue auction at the same time.

larrythecucumber321:
Ahh understood. I think this flow should be good then. I just think of the Register button as another MEV searcher progressing system state 🙂

contracts/facade/FacadeAct.sol
```diff
@@ -351,4 +351,98 @@ contract FacadeAct is IFacadeAct {
...
+                bm.settleTrade(erc20s[i]);
+            }
+        }
+
+        uint256 tradesOpen = bm.tradesOpen();
+        if (tradesOpen != 0) return false;
+
+        // Try to launch auctions
+        bm.manageTokensSortedOrder(new IERC20[](0));
+        return bm.tradesOpen() - tradesOpen > 0;
```
pmckelvy1, line 378: no need to subtract `tradesOpen` since it has to be 0 here

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+        await aToken.setRewards(backingManager.address, rewardAmountAAVE)
 
-      // Run auction via Facade
-      let [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
-      expect(addr).to.equal(backingManager.address)
-      expect(data).to.not.equal('0x')
+        // Via Facade get next call - will claim rewards from backingManager
+        let [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 377: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+        const sellAmtRToken: BigNumber = rewardAmountAAVE.sub(sellAmt) // Remainder
+        const minBuyAmtRToken: BigNumber = await toMinBuyAmt(sellAmtRToken, fp('1'), fp('1'))
 
-      // End current auction
-      await expect(
-        owner.sendTransaction({
+        // Via Facade get next call - will transfer RToken to Trader
+        ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 396: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+          .withArgs(anyValue, aaveToken.address, rsr.address, sellAmt, minBuyAmt)
 
-      // Via Facade get next call - will claim rewards from backingManager
-      let [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
-      expect(addr).to.equal(backingManager.address)
-      expect(data).to.not.equal('0x')
+        // Check no new calls to make from Facade
+        ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(ZERO_ADDRESS)
+        expect(data).to.equal('0x')
```
pmckelvy1, line 493: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+        await backingManager.claimRewards()
 
-      // Next call would start Revenue auction - RTokenTrader
-      ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
-      expect(addr).to.equal(rTokenTrader.address)
-      expect(data).to.not.equal('0x')
+        // Via Facade get next call - will transfer RSR to Trader
+        ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 518: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+        await aToken.setRewards(backingManager.address, rewardAmountAAVE)
 
-      // Via Facade get next call - will transfer RSR to Trader
-      ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
-      expect(addr).to.equal(backingManager.address)
-      expect(data).to.not.equal('0x')
+        // Via Facade get next call - will not attempt to claim - No action taken
+        const [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(ZERO_ADDRESS)
+        expect(data).to.equal('0x')
```
pmckelvy1, line 595: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
-        owner.sendTransaction({
+        const rewardAmountAAVE = bn('0.5e18')
+
+        // AAVE Rewards
+        await aToken.setRewards(backingManager.address, rewardAmountAAVE)
+
+        // Via Facade get next call - will claim rewards from backingManager
+        let [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 614: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
-        )
-
-      // Via Facade get next call - will open RSR trade
-      ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
-      expect(addr).to.equal(rsrTrader.address)
-      expect(data).to.not.equal('0x')
+        // Via Facade get next call - will Claim and sweep rewards
+        const [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 767: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
-      // Close auction in RSR Trader
-      await expect(
-        owner.sendTransaction({
+        // AAVE Rewards - RSR Trader
+        await aToken.setRewards(rsrTrader.address, rewardAmountAAVE)
+
+        // Via Facade get next call - will claim rewards from Traders, via Facade
+        let [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(facadeAct.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 787: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+        let [addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(ZERO_ADDRESS)
+        expect(data).to.equal('0x')
+
+        // RSR can be distributed with no issues - seed stRSR with half as much
+        await rsr.connect(addr1).transfer(backingManager.address, hndAmt)
+        await rsr.connect(addr1).transfer(stRSR.address, hndAmt.div(2))
+        ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 847: check function signature

test/FacadeAct.test.ts
```diff
@@ -200,669 +203,739 @@ describe('FacadeAct contract', () => {
...
+        expect(data).to.equal('0x')
 
-      // Check rewards collected
-      expect(await aaveToken.balanceOf(rsrTrader.address)).to.equal(rewardAmountAAVE)
-      expect(await aaveToken.balanceOf(rTokenTrader.address)).to.equal(bn(0))
+        // RToken can be distributed with no issues
+        await rToken.connect(addr1).transfer(backingManager.address, hndAmt)
+        ;[addr, data] = await facadeAct.callStatic.getActCalldata(rToken.address)
+        expect(addr).to.equal(backingManager.address)
+        expect(data).to.not.equal('0x')
```
pmckelvy1, line 879: check function signature

pmckelvy1 (changes requested):
there are a handful of places where we are implicitly checking the returned call data / function signature by expecting an event from the call, but there are places where we don't do this, so we should explicitly check that the returned data has the expected function signature.

test/FacadeAct.test.ts
tbrent: I agree that these tests would be better if they also checked the function signature, but `getActCalldata` is not part of the changes in this PR.

tbrent: I didn't add these tests in this PR, I just reorg'd the test file and it made it look like it was all new code.

pmckelvy1: ah ok. i'll make an issue to touch these up though as it seems valuable to test

PR #728, opened 2023-03-24
expose StRSR.draftRSR/stakeRSR/totalDrafts

### LINKED ISSUE #727: Expose `draftRSR` + `stakeRSR` on StRSR
There is currently no way to know how much RSR is in the withdrawal queue. We should expose both these internal vars.

tbrent:
Turns out it's not possible to view the amount of RSR (or StRSR) being withdrawn. We definitely want to expose this. I've implemented it in a way that obviously safe and doesn't touch storage. `StRSRP1Votes` remains well under the contract limit. 

Closes #727

### DIFF
--- a/contracts/p1/StRSR.sol
+++ b/contracts/p1/StRSR.sol
@@ -465,6 +465,21 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
         return draftQueues[era_][account].length;
     }
 
+    /// @return {qDrafts} The amount of RSR currently being withdrawn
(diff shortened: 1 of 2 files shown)


tbrent:
Well that's good to find out...reordering doesn't fix the issue.

julianmrodri (approved):
LGTM.  We needed this! So if we wanted to know the total stRSR unstaked but not yet withdrawn that would be `totalDrafts()` and thats it?

tbrent:
> LGTM. We needed this! So if we wanted to know the total stRSR unstaked but not yet withdrawn that would be `totalDrafts()` and thats it?

Yep!

PR #729, opened 2023-03-24
give collateral plugins 3 oracles by default.

pmckelvy1:
This pr adds the concept of 2 additional oracles to ALL collaterals that inherit FiatCollateral.

The reason for this is to allow for testing (namely, integration scripts and fork testing scripts like the WIP upgrade-checker.ts) to programmatically push any and all oracles forward so that collaterals do not become `DISABLED` when the timestamp of the chain gets pushed forward to do things like pass governance proposals or realize revenue.


pmckelvy1:
closing for now, need to think on this more.

PR #731, opened 2023-03-24
fix bad comments in LidoStakedEthCollateral plugin

tbrent:


tbrent:
Look like @pmckelvy1 has this covered in #729

julianmrodri (approved):
lgtm

PR #732, opened 2023-03-26
Update collateral asset plugin list

larrythecucumber321:
Updated based on deployments in `scripts/addresses/mainnet-2.1.0/1-tmp-assets-collateral.json`


tbrent:
LGTM!

PR #746, opened 2023-03-29
MIM/3Pool tests+scripts

tbrent:
I can't believe thomas got me to do this


contracts/plugins/assets/convex/CvxStableMetapoolCollateral.sol
```diff
@@ -91,13 +82,41 @@ contract CvxStableMetapoolCollateral is CvxStableCollateral {
         return (low, high, 0);
     }
 
+    /// Can revert, used by `_anyDepeggedOutsidePool()`
+    /// Should not return FIX_MAX for low
+    /// Should only return FIX_MAX for high if low is 0
+    /// @return low {UoA/tok} The low price estimate
+    /// @return high {UoA/tok} The high price estimate
+    function tryPairedPrice() public view virtual returns (uint192 low, uint192 high) {
+        uint192 p = chainlinkFeed.price(oracleTimeout); // {UoA/tok}
```
pmckelvy1, line 91: not sure what the units should be here but i don't think they're `{UoA/tok}`. convex pool units seem to be even more abstract than our already 3-level-deep abstractions...

contracts/plugins/assets/convex/CvxStableMetapoolCollateral.sol
```diff
@@ -91,13 +82,41 @@ contract CvxStableMetapoolCollateral is CvxStableCollateral {
...
+    }
+
     // === Internal ===
 
     /// @return {ref/tok} Actual quantity of whole reference units per whole collateral tokens
     function _underlyingRefPerTok() internal view override returns (uint192) {
         return _safeWrap(metapool.get_virtual_price());
     }
 
+    // Override this in child classes to implement metapools
```
pmckelvy1, line 103: delete

contracts/plugins/assets/convex/CvxStableRTokenMetapoolCollateral.sol
```diff
@@ -82,4 +82,9 @@ contract CvxStableRTokenMetapoolCollateral is CvxStableMetapoolCollateral {
         rTokenOracle.price(IRToken(address(pairedToken)), false); // refresh price in oracle
         super.refresh();
     }
+
+    // Override CvxStableMetapoolCollateral, since paired RToken cannot default
+    function _anyDepeggedOutsidePool() internal view virtual override returns (bool) {
+        return false;
```
pmckelvy1, line 88: why not check the RTokenPriceOracle here?

pmckelvy1 (changes requested):

contracts/plugins/assets/convex/CvxStableMetapoolCollateral.sol
akshatmittal: I think `{UoA/tok}` is correct especially since this is the stable collateral. (And I agree about the Convex/Curve units, damn)

contracts/plugins/assets/convex/CvxStableRTokenMetapoolCollateral.sol
akshatmittal: I don't think we check for depegged state for RTokens, but maybe we should.

contracts/plugins/assets/convex/CvxStableMetapoolCollateral.sol
tbrent: Good point, these should be `{UoA/pairedTok}`

contracts/plugins/assets/convex/CvxStableRTokenMetapoolCollateral.sol
tbrent: Added

contracts/plugins/assets/convex/CvxStableRTokenMetapoolCollateral.sol
```diff
@@ -75,16 +75,36 @@ contract CvxStableRTokenMetapoolCollateral is CvxStableMetapoolCollateral {
...
+    function tryPairedPrice()
+        public
+        view
+        virtual
+        override
+        returns (uint192 lowPaired, uint192 highPaired)
+    {
+        // refresh price in oracle if needed
+        (Price memory p, uint48 ts) = rTokenOracle.priceView(IRToken(address(pairedToken)));
+        require(block.timestamp - ts <= rTokenOracle.cacheTimeout(), "call refresh()");
```
pmckelvy1, line 92: maybe we move this responsibility into the RTokenPriceOracle and give the function a parameter like `allowStalePrice`? just a thought bc i think we'll end up needing this check in a lot of places and it would be a shame to forget it

pmckelvy1 (changes requested):

contracts/plugins/assets/convex/CvxStableRTokenMetapoolCollateral.sol
pmckelvy1: yup that makes sense

test/plugins/individual-collateral/convex/constants.ts
```diff
@@ -72,7 +74,7 @@ export const FRAX_BP = '0xDcEF968d416a41Cdac0ED8702fAC8128A64241A2'
 export const FRAX_BP_TOKEN = '0x3175Df0976dFA876431C2E9eE6Bc45b65d3473CC'
 
 // eUSD + fraxBP -- this metapool combines lpToken + curvePool
-export const RTOKEN_ORACLE = '0x16fc605eEE1ece04792b563586ad530C9d535054'
+export const RTOKEN_ORACLE = '0x33B2a7216E3fC4eA33D8A69Ed80BA65d626eaD50'
```
pmckelvy1, line 77: did you redeploy this already?

pmckelvy1 (changes requested):

contracts/plugins/assets/convex/CvxStableRTokenMetapoolCollateral.sol
akshatmittal: Cascades nicely into the RToken Oracle conversation (yes I keep bringing it up, but it reminds me of that haha)

PR #759, opened 2023-04-05
Facade read hotfix

tbrent:
The use of high prices in `FacadeRead.basketBreakdown` can cause some weird behavior when a registered asset is UNPRICED.

- Change `basketBreakdown`/`backingOverview` to rely on low prices only
- Add consideration of backingBuffer to `traderBalances`


julianmrodri (approved):
LGTM

PR #760, opened 2023-04-06
Cvx eusd price

tbrent:
- Make `tryPairedPrice` accept a stale price, if refresh() is either failing or the price is being read out from a Facade or something
- Invert `priceView` boolean to make `price`/`priceView` parallel to each other
- Move RTokenOracle out of a constants file and into our master address configuration file


tbrent:
Still have to deploy a new `cvxeUSDFRAXBP` address, if we confirm this direction is correct.

tbrent:
Closing. Will create a new PR that removes the RTokenOracle contract entirely.

PR #761, opened 2023-04-07
Gitcoin Morpho AAVE Bounty PR

mattjurenka:


jankjr:
Greetings @mattjurenka

I've rebased your code with v3.0.0 and converted the unit test suite into our generic plugin suite for all three variants of the tokens.

You can follow the progress here https://github.com/reserve-protocol/protocol/pull/853, feel free to comment if I missed anything some invariant you were testing in the original spec, or I messed something up.

PR #762, opened 2023-04-07
Cheaper manage tokens

tbrent:
- Prevent `basketRange()` from being re-executed
- Avoid recalculation of the price of a basket
- Avoid re-lookup of basket quantities

I tried caching prices like the quantities are cached as well, but it turned out to be worse in most cases. 

Also note that the requirement of the `.prices()` function on the RTokenAsset makes this an interface breaking change and therefore requires it goes in the 3.0.0 release.


tbrent:
reviewer should read through `RTokenAsset.sol` carefully and ensure the try/catch's make sense. previous version only had them for `price()`.

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -391,9 +414,36 @@ library RecollateralizationLibP1 {
...
 
+        // Special-case RToken
+        {
+            RTokenAsset rTokenAsset = RTokenAsset(
+                address(ctx.ar.toAsset(IERC20(address(ctx.rToken))))
+            );
+            uint192 bal = rTokenAsset.bal(address(ctx.bm));
+
+            if (
+                ntpCtx.rTokenPrice.high > 0 &&
```
pmckelvy1, line 425: why use rTokenPrice.high here but rTokenLotPrice.low below?

contracts/plugins/assets/RTokenAsset.sol
```diff
@@ -72,9 +135,8 @@ contract RTokenAsset is IAsset {
     /// @return {UoA/tok} The lower end of the price estimate
     /// @return {UoA/tok} The upper end of the price estimate
     function price() public view virtual returns (uint192, uint192) {
-        try this.tryPrice() returns (uint192 low, uint192 high) {
-            assert(low <= high);
-            return (low, high);
+        try this.prices() returns (Price memory p, Price memory) {
```
pmckelvy1, line 138: seems like price() and lotPrice() are now way more inefficient?  necessary unfortunate side-effect or can we still have tryPrice() for some efficiency?

pmckelvy1 (changes requested):

contracts/p1/BasketHandler.sol
```diff
@@ -333,45 +333,53 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
...
         for (uint256 i = 0; i < len; ++i) {
             uint192 qty = quantity(basket.erc20s[i]);
             if (qty == 0) continue;
 
-            (uint192 lowP, uint192 highP) = useLotPrice
-                ? assetRegistry.toAsset(basket.erc20s[i]).lotPrice()
-                : assetRegistry.toAsset(basket.erc20s[i]).price();
+            IAsset asset = assetRegistry.toAsset(basket.erc20s[i]);
+            (uint192 lowP, uint192 highP) = asset.price();
+            (uint192 lotLowP, uint192 lotHighP) = asset.price();
```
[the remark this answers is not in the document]
tbrent: Good catch! Yes, indeed, lotPrice. And no, we're not currently testing this, the tests pass either way. I'll write a regression test...

contracts/plugins/assets/RTokenAsset.sol
tbrent: They're definitely more inefficient. My thought process was that `RTokenAsset.price()` only has a single use left: `RevenueTrader.manageToken()`. We don't call it from BackingManager anymore, and since RTokenAsset doesn't have price checks in `refresh()` we don't call it during `assetRegistry.refresh()`. 

But I think this logic fails in two ways:
1. Collateral that use the RTokenAsset price, such as `CvxStableRTokenMetapoolCollateral`,  _does_ call the price during refresh(). 
2. It also bundles the revert behavior of `price()`/`lotPrice()` together, which we don't want. Very possible for the underlying collateral's `price()` to revert and for `lotPrice()` not to. We should maintain that property here.

All to say, adding `tryPrice()` back...

contracts/p1/mixins/RecollateralizationLib.sol
tbrent: This is as it was before, and I believe it's correct. It's for the case where the RToken price is unknown ([0, FIX_MAX]). In that case it's correct to use 0 as the sellPrice and allow the auction to handle price discovery, as long as the `isEnoughToSell()` call says the low lotPrice is sufficiently large.

Now, you could argue the `high > 0` check here is redundant with the `isEnoughToSell` call...The only case I can see where the `high > 0` check would differentiate itself would be if there were a mistake in the lotPrice, such that price() were [0, 0] but lotPrice() were nonzero.

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -391,9 +414,36 @@ library RecollateralizationLibP1 {
             }
         }
 
+        // Special-case RToken
+        {
```
pmckelvy1, line 418: just leaving a comment to remind myself to review this deeper in the morning when i'm fresh. it seems ok but but its also definitely a rather large change so i want to be sure it matches the previous logic.

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -391,9 +414,36 @@ library RecollateralizationLibP1 {
...
+            RTokenAsset rTokenAsset = RTokenAsset(
+                address(ctx.ar.toAsset(IERC20(address(ctx.rToken))))
+            );
+            uint192 bal = rTokenAsset.bal(address(ctx.bm));
+
+            if (
+                ntpCtx.rTokenPrice.high > 0 &&
+                isBetterSurplus(
+                    maxes,
+                    CollateralStatus.SOUND,
```
pmckelvy1, line 428: why are we making the assumption that the rtoken is sound? it matches the previous logic but im just wondering if that is correct here

tbrent: The `RTokenAsset' isn't a collateral, it's just an asset. In the loop above you can see that all the other assets are also written to essentially have a SOUND collateral status for purposes of comparison.

Also, at the top of `manageTokens()` there's a require that the basket is SOUND.

PR #767, opened 2023-04-10
Feat/warmup period

julianmrodri:
* Add warmup period logic 
* Adapt contracts and tests
* Adds minWarmup period and adapts prior tests.
* Revert on trading delay
* Adds 2 non-related minor things:  
   - Comments in `IBackingManger` that were missing
   - `describeFork` in StaticAToken which should only run on Fork (this was not being enforced).

Notes:
 - Because now we have to use `TestIBasketHandler` and not `IBasketHandler` a lot of tests files just have this simple change. It is now more consistent with all the other components.
 - The new variables are added to the end to prevent stepping on previous storage Slots.  The 3 variables are being packed in one slot, so we only need to decrease `gap` by 1.

### DIFF
--- a/contracts/interfaces/IBasketHandler.sol
+++ b/contracts/interfaces/IBasketHandler.sol
@@ -82,14 +82,15 @@ interface IBasketHandler is IComponent {
     /// @custom:interaction
     function refreshBasket() external;
 
+    /// Track the basket status changes
+    /// @custom:refresher
+    function trackStatus() external;
+
     /// @return If the BackingManager has sufficient collateral to redeem the entire RToken supply
     function fullyCollateralized() external view returns (bool);
 
     /// @return status The worst CollateralStatus of all collateral in the basket
     function status() external view returns (CollateralStatus status);
-
-    /// Track the basket status changes
-    function trackStatus() external;
 
     /// @return If the basket is ready to issue and trade
     function isReady() external view returns (bool);
--- a/contracts/interfaces/IBasketHandler.sol
+++ b/contracts/interfaces/IBasketHandler.sol
@@ -38,8 +38,21 @@ interface IBasketHandler is IComponent {
     /// @param erc20s The set of backup collateral tokens
     event BackupConfigSet(bytes32 indexed targetName, uint256 indexed max, IERC20[] erc20s);
 
+    /// Emitted when the warmup period is changed
+    /// @param oldVal The old warmup period
+    /// @param newVal The new warmup period
+    event WarmupPeriodSet(uint48 indexed oldVal, uint48 indexed newVal);
+
+    /// Emitted when the status of a basket has changed
+    /// @param oldStatus The previous basket status
+    /// @param newStatus The new basket status
+    event BasketStatusChanged(
+        CollateralStatus indexed oldStatus,
+        CollateralStatus indexed newStatus
+    );
+
     // Initialization
-    function init(IMain main_) external;
+    function init(IMain main_, uint48 warmupPeriod_) external;
 
     /// Set the prime basket
     /// @param erc20s The collateral tokens for the new prime basket
@@ -74,6 +87,13 @@ interface IBasketHandler is IComponent {
 
     /// @return status The worst CollateralStatus of all collateral in the basket
     function status() external view returns (CollateralStatus status);
+
+    /// Track the basket status changes
+    /// @custom:protected
+    function trackStatus() external;
+
+    /// @return If the basket is ready to issue and trade
+    function isReady() external view returns (bool);
 
     /// @param erc20 The ERC20 token contract for the asset
     /// @return {tok/BU} The whole token quantity of token in the reference basket
@@ -119,4 +139,10 @@ interface IBasketHandler is IComponent {
 
     /// @return The current basket nonce, regardless of status
     function nonce() external view returns (uint48);
+}
+
+interface TestIBasketHandler is IBasketHandler {
+    function warmupPeriod() external view returns (uint48);
+
+    function setWarmupPeriod(uint48 val) external;
 }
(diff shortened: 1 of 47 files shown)


contracts/p0/BasketHandler.sol
```diff
@@ -671,5 +715,5 @@ contract BasketHandlerP0 is ComponentP0, IBasketHandler {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[48] private __gap;
+    uint256[47] private __gap;
```
tbrent, line 718: We can honestly get rid of this gap in the p0 contracts

contracts/p1/BasketHandler.sol
```diff
@@ -138,6 +139,19 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
     // and everything except redemption should be paused.
     bool private disabled;
 
+    // These are effectively local variables of _switchBasket.
+    // Nothing should use their values from previous transactions.
+    EnumerableSet.Bytes32Set private _targetNames;
+    Basket private _newBasket; // Always empty
```
tbrent, line 145: Good call to move these up. I was adding my new variables down at the bottom by the storage gap, but this will make it easier to track all variables and make sure we don't accidentally insert a new variable in the middle.

contracts/p1/BasketHandler.sol
```diff
@@ -764,5 +808,5 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[42] private __gap;
+    uint256[41] private __gap;
```
tbrent, line 811: Did you check this? The 2 uint48s and the CollateralStatus enum all fit in one slot?

test/Main.test.ts
```diff
@@ -1739,7 +1830,7 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
       expect(await facadeTest.callStatic.totalAssetValue(rToken.address)).to.equal(0)
     })
 
-    it('Should handle full collateral deregistration and disable the basket', async () => {
+    it.skip('Should handle full collateral deregistration and disable the basket', async () => {
```
tbrent, line 1833: Should this be skipped?

tbrent:
Left some comments, really nice though!

tbrent:
LGTM to merge after resolving comments

contracts/p0/BasketHandler.sol
julianmrodri: Yes we should probably do that to avoid confusions.

contracts/p1/BasketHandler.sol
julianmrodri: I think its safer and its kind of the norm to have all variables on top. Felt weird to have them at the bottom. I think when people read contracts you do not expect variables at the bottom. 👍

contracts/p1/BasketHandler.sol
julianmrodri: Yep.. totally checked

julianmrodri: Yes, checked.  Its packing everything in one slot because we only have few values in that enum. We have room for 160 values we could store in that slot so I think we are ok.

julianmrodri: But it is true enums could take up to a full slot, its not the case for this enum so I think we are safe.

contracts/p1/BackingManager.sol
```diff
@@ -107,8 +107,9 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
 
         if (tradesOpen > 0) return;
-        // Only trade when all the collateral assets in the basket are SOUND
-        require(basketHandler.status() == CollateralStatus.SOUND, "basket not sound");
+
+        // Ensure basket is ready, SOUND and not in warmup period
+        require(basketHandler.isReady(), "basket not ready");
 
         uint48 basketTimestamp = basketHandler.timestamp();
         if (block.timestamp < basketTimestamp + tradingDelay) return;
```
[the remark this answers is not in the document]
julianmrodri: Im ok with changing this. Im not sure I fully understand the change on `runAuctions...`. I think if we change it,. it would be cleaner to simply adjust the tests but let `runAuctions...` do the call to `manageTokens`as usual.

contracts/p1/BasketHandler.sol
tbrent: Sweet. I'm gonna end up packing another `uint48` in that storage slot in #771

julianmrodri:
Implemented all the changes, including the minimum warmup period and reverting on the trading delay.

contracts/interfaces/IBasketHandler.sol
```diff
@@ -75,6 +88,13 @@ interface IBasketHandler is IComponent {
     /// @return status The worst CollateralStatus of all collateral in the basket
     function status() external view returns (CollateralStatus status);
 
+    /// Track the basket status changes
+    /// @custom:protected
```
tbrent, line 92: This is not actually protected anymore, correct?

tbrent:
Left one further comment, but good to merge after that!

tbrent:
Actually, I'm just going to push this through now

tbrent:
@julianmrodri I moved the `trackStatus()` function up higher next to `refreshBasket()` and added the `@custom:refresher` tag

contracts/interfaces/IBasketHandler.sol
julianmrodri: Missed this one. Will fix.

PR #769, opened 2023-04-11
Ability to cancel unstaking request

akshatmittal:
Cancelling an unstaking does not require waiting for the entire duration of the withdrawal delay.


contracts/p1/StRSR.sol
```diff
@@ -335,6 +335,54 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
...
 
+    function cancelUnstake(uint256 endId) external notPausedOrFrozen {
+        address account = _msgSender();
+
+        // == Refresh ==
+        assetRegistry.refresh();
+
+        // == Checks + Effects ==
+        require(basketHandler.fullyCollateralized(), "RToken uncollateralized");
+        require(basketHandler.status() == CollateralStatus.SOUND, "basket defaulted");
```
akshatmittal, line 346: @tbrent I'm contemplating removing these two checks given that we're not removing RSR from the system. What do you think?

tbrent: Absolutely. You _should_ be able to re-stake when undercollateralized, or when basket is not sound.

tbrent: We can also remove the assetRegistry refresh, I see no reason to need that

contracts/p1/StRSR.sol
```diff
@@ -335,6 +335,54 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
...
+
+        uint256 firstId = firstRemainingDraft[draftEra][account];
+        CumulativeDraft[] storage queue = draftQueues[draftEra][account];
+        if (endId == 0 || firstId >= endId) return;
```
tbrent, line 350: So, just because `withdraw()` does this doesn't mean `cancelUnstake()` has to. I do prefer the revert behavior myself. 

I would also check to see what it breaks in tests to use revert in `withdraw()` as well. If it's not much, maybe it's worth doing.

contracts/p1/StRSR.sol
akshatmittal: Yeah I think you're right about the refresh. I only kept it if under collateralized was a factor. Removed it.

tbrent:
Other than the comments I left, the only other thing I would contemplate is whether `restake` is better than `cancelUnstaking`. Either way, though. 

LGTM otherwise! Let me know what you find out about the `withdraw()` revert, that seems like the only thing left at this point that could conceivably add to this PR.

akshatmittal:
> I would contemplate is whether `restake` is better than `cancelUnstaking`.

Yeah I thought about it and ended up on `cancelUnstaking` mainly because I wasn't able to justify what `restake` even means, plus we aren't doing partials here. For all intensive purposes, we're cancelling a request to unstake at best. Thoughts?

tbrent:
> > I would contemplate is whether `restake` is better than `cancelUnstaking`.
> 
> Yeah I thought about it and ended up on `cancelUnstaking` mainly because I wasn't able to justify what `restake` even means, plus we aren't doing partials here. For all intensive purposes, we're cancelling a request to unstake at best. Thoughts?

Good point...the atoms of the action make it less like re-staking and more like cancelling an unstake. 👍

PR #771, opened 2023-04-11
Multiple nonce redemption

### LINKED ISSUE #795: redeem at previous basket nonces
Can allow redeemer to redeem at any basket, even a previously defaulted one, as long as the redemption is prorata based on balances held by the BackingManager. Might even be able to do a linear combination of basket nonces. 

Requires changes to the BasketHandler's storage to track all previous baskets. 

To think about: How do reward balances that could be claimed factor in? Can you get more out of the system than you should be able to if reward balances have accumulated to one of the collateral tokens being swapped out, mid-rebalancing?

tbrent:
Proposed new redemption semantics

Old functions `redeem(.., uint48 basketNonce)` and `redeemTo(.., uint48 basketNonce)`:
- remove `basketNonce` argument; always implicitly on latest nonce
- revert if partial redemption in any way; ie undercollateralized

New function `redeemCustom(...)`
- accept an array of basket nonces and proportions
- as well as minimum token amounts out
- only redemption method possible when undercollateralized

Resolves #795


contracts/p1/RToken.sol
```diff
@@ -178,79 +182,159 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
+    /// @param minAmounts {qTok} The minimum ERC20 quantities the caller should receive
+    /// @custom:interaction RCEI
+    function redeemToCustom(
         address recipient,
         uint256 amount,
-        uint48 basketNonce
-    ) public notFrozen exchangeRateIsValidAfter {
+        uint48[] memory basketNonces,
+        uint192[] memory portions,
+        IERC20[] memory erc20sOut,
```
pmckelvy1, line 291: i'm thinking we should validate uniqueness in this array

contracts/p1/RToken.sol
```diff
@@ -178,79 +182,159 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
+    /// @param portions {1} An array of Fix quantities that must add up to FIX_ONE
+    /// @param erc20sOut An array of ERC20s expected out
+    /// @param minAmounts {qTok} The minimum ERC20 quantities the caller should receive
+    /// @custom:interaction RCEI
+    function redeemToCustom(
         address recipient,
         uint256 amount,
-        uint48 basketNonce
-    ) public notFrozen exchangeRateIsValidAfter {
+        uint48[] memory basketNonces,
```
pmckelvy1, line 289: validate uniqueness here as well?

contracts/p1/RToken.sol
```diff
@@ -178,79 +182,159 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
+        emit BasketsNeededChanged(basketsNeeded, basketsNeeded - basketsRedeemed);
+        basketsNeeded = basketsNeeded - basketsRedeemed;
 
-        require(basketHandler.nonce() == basketNonce, "non-current basket nonce");
-        (address[] memory erc20s, uint256[] memory amounts) = basketHandler.quote(
-            basketsRedeemed,
-            FLOOR
+        // === Get basket redemption amounts ===
+
+        (address[] memory erc20s, uint256[] memory amounts) = basketHandler.quoteCustomRedemption(
```
pmckelvy1, line 326: maybe rename to `quotedERC20s` and `quotedAmounts` for clarity

contracts/p1/RToken.sol
pmckelvy1: looking through this again, i don't think non-uniqueness could be malicious, but it could waste gas...though checking uniqueness costs gas...not sure whats best here. a little extra gas to err on the side of caution?

contracts/p1/BasketHandler.sol
```diff
@@ -440,12 +460,95 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
...
+
+    /// Return the redemption value of `amount` BUs for a linear combination of historical baskets
+    /// @param basketNonces An array of basket nonces to do redemption from
+    /// @param portions {1} An array of Fix quantities
+    /// @param amount {BU}
+    /// @return erc20s The backing collateral erc20s
+    /// @return quantities {qTok} ERC20 token quantities equal to `amount` BUs
+    // Returns (erc20s, [quantity(e) * amount {as qTok} for e in erc20s])
+    function quoteCustomRedemption(
+        uint48[] memory basketNonces,
```
pmckelvy1, line 483: we should write some tests to ensure non-uniqueness cannot be used maliciously. i don't think it can be but it is not immediately obvious

test/Main.test.ts
```diff
@@ -1715,6 +1725,357 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
         .withArgs([token0.address], [fp('1')], [ethers.utils.formatBytes32String('USD')])
     })
 
+    describe.only('Historical Redemptions', () => {
```
pmckelvy1, line 1728: remove only

pmckelvy1 (changes requested):

contracts/p1/RToken.sol
tbrent: Seems fine? The user can provide as many duplicate entries as they want and it doesn't break anything, just costs them more gas. Given protecting against this would involve spending gas _always_, I feel like the tradeoff is not worth it.

contracts/p1/RToken.sol
tbrent: Seems fine to me as well. The same basket nonce ten times with portions of 10% each gives the same result as providing the nonce just once and using a portion of 100%.

contracts/p1/RToken.sol
tbrent: I also think I want to return the amounts transferred, so this can be used in simulation easily.

contracts/p1/BasketHandler.sol
tbrent: Agree we should have a test for it; I'll add it. I do think it's safe, but we should have the test anyway.

test/Main.test.ts
tbrent: Yup; I hadn't made a pass on the tests yet for the record

contracts/p1/RToken.sol
```diff
@@ -178,79 +182,159 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
-    ) public notFrozen exchangeRateIsValidAfter {
+        uint48[] memory basketNonces,
+        uint192[] memory portions,
+        IERC20[] memory erc20sOut,
+        uint256[] memory minAmounts
+    ) external notFrozen exchangeRateIsValidAfter {
         // == Refresh ==
+
         assetRegistry.refresh();
+        main.furnace().melt();
```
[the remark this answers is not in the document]
tbrent: So, this one is interesting. Previously we had a try/catch because the modifier on the Furnace was `notPausedOrFrozen`, while redeem() was `notFrozen`. Now the modifier on the Furnace is `notFrozen`, because melting is required (i) in issuance, and (ii) during trading. So it _should_ be the case that we don't need a try/catch. But we're also talking redemption here! It would be really bad, for example, to be able to brick redemption by swapping out the Furnace maliciously. So I think I'll throw the try/catch back in there, for safety. Good catch.

contracts/p1/RToken.sol
pmckelvy1: yeah i think the only thing we want to prove is that non-uniqueness cannot be malicious. as long as we can do that, no need to check in the call

contracts/p1/RToken.sol
pmckelvy1: yeah i think its fine, worth writing a test or two for it though

PR #777, opened 2023-04-14
Dutch Auctions

tbrent:
Implementation only, so far. 

How it works:
- Both Traders track dutch auctions by a `tradeEnd` variable, which is the ending timestamp of _either_ the most recent dutch auction or batch auction
  - If `tradeEnd` is in the future by less than or equal to `dutchAuctionLength`, then an auction is currently ongoing and it is already reflected in storage
  - If `tradeEnd` is in the past by less than `dutchAuctionLength`, then an auction is currently ongoing but not yet reflected in storage
  - This structure allows the protocol to forego the starting tx for a dutch auction and loop until nobody is interested in bidding anymore
- `tradeEnd` gets pushed forward by 1 `dutchAuctionLength` by the first bid in an auction
- The only other way `tradeEnd` can change is that it can be brought up to the present, and this has very specific conditions for each trader
  - BackingManager: After a reference basket change (+ additional delays)
  - RevenueTrader: Whenever the BackingManager does `handoutExcessAssets`
- `tradeEnd` cannot go backwards in time

In this way, nobody has to start rebalancing txs, they just happen autonomously at the hand of MEV searchers after a basket change. On the RevenueTrader side the dutch auctions are triggered after `BackingManager.handoutExcessAssets()`. 

Batch auctions can only ever be _downstream_ of dutch auctions, and when a dutch auction fails to get any bids it cannot be restarted (except in the RevenueTrader if someone calls `BackingManager.handoutExcessAssets()` again). 

Bugfixes
- Trading takes into account `maxTradeVolume` contribution from both sides


contracts/p1/mixins/Trading.sol
```diff
@@ -28,10 +31,24 @@ abstract contract TradingP1 is Multicall, ComponentP1, ReentrancyGuardUpgradeabl
...
+
+    // {s} the length of the implicit falling-price dutch auction
+    uint48 public dutchAuctionLength;
+
+    // At the start of a tx, tradeEnd can be:
+    //   1. more than dutchAuctionLength away => No dutch auction ongoing
+    //   2. within dutchAuctionLength in the past => Dutch auction with 0 bids ongoing
+    //   3. within dutchAuctionLength in the future => Dutch auction with 1+ bids ongoing
+    // [X, Y): inclusive on the left-bound and exclusive on the right-bound
+    uint48 internal tradeEnd; // {s} timestamp of the end of the last trade (batch OR dutch)
```
pmckelvy1, line 48: i think we want to clean up the usage of the word "trade". on line 225 we distinguish that a dutch-auction swap is not a trade https://github.com/reserve-protocol/protocol/pull/777/files#diff-b3b12b74962f08fc087a7944168db17cebbd1f7c989fe73a7b047957e6a450d4R225 but we also seem to be using this `tradeEnd` variable to time a dutch-auction swap.  maybe `lastAuctionEnd` is a better name here?

contracts/p1/BackingManager.sol
```diff
@@ -146,14 +130,106 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+        uint256 amountOut
+    ) external returns (Swap memory s) {
+        // == Refresh ==
+        assetRegistry.refresh();
+        furnace.melt();
+
+        requireCanTrade(0);
+
+        // executeSwap if storage auction already exists
+        if (tradeEnd > block.timestamp) {
```
pmckelvy1, line 160: since we do this check in multiple places, lets just make a descriptive function for it.  `dutchAuctionExists()` or something like that

contracts/p1/BackingManager.sol
```diff
@@ -146,14 +130,106 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+            return executeSwap(dutchAuctions[tradeEnd], amountOut);
+        }
+
+        // === Checks/Effects ===
+
+        require(tradeEnd + dutchAuctionLength > block.timestamp, "no dutch auction ongoing");
```
pmckelvy1, line 166: same here. `dutchAuctionActive()`?

contracts/p1/BackingManager.sol
```diff
@@ -146,14 +130,106 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+        tradeEnd += dutchAuctionLength;
+
+        // Same TradeRequest from manageTokens()
+        (bool doTrade, TradeRequest memory req) = RecollateralizationLibP1
+            .prepareRecollateralizationTrade(this, basketHandler.basketsHeldBy(address(this)));
+
+        require(doTrade, "swap not available");
+        require(tokenIn == req.buy.erc20() && tokenOut == req.sell.erc20(), "ERC20 swap expired");
```
pmckelvy1, line 174: not sure this is the right error message? "Swap tokens mismatch" might make more sense?

contracts/p1/BackingManager.sol
```diff
@@ -146,14 +130,106 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+
+    /// @return The ongoing auction as a Swap
+    function getDutchAuctionQuote() external view returns (Swap memory) {
+        requireCanTrade(0);
+
+        if (tradeEnd > block.timestamp) {
+            return dutchAuctions[tradeEnd].toSwap(progression());
+        }
+
+        require(tradeEnd + dutchAuctionLength > block.timestamp, "no dutch auction ongoing");
```
pmckelvy1, line 202: do we want to revert on view functions, or is it better to just return an empty swap? or we can return a swap w/ a bool like we do for `prepareRecollateralizationTrade`?

contracts/p1/BackingManager.sol
```diff
@@ -146,14 +130,106 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+
+        // Same TradeRequest from manageTokens()
+        (bool doTrade, TradeRequest memory req) = RecollateralizationLibP1
+            .prepareRecollateralizationTrade(this, basketHandler.basketsHeldBy(address(this)));
+
+        require(doTrade, "swap not available");
```
pmckelvy1, line 208: same q as above

contracts/p1/RevenueTrader.sol
```diff
@@ -89,14 +119,121 @@ contract RevenueTraderP1 is TradingP1, IRevenueTrader {
...
+
+        require(address(trades[tokenOut]) == address(0), "nonatomic trade ongoing");
+        require(tokenIn == tokenToBuy, "will only buy tokenToBuy");
+        require(tokenOut != tokenToBuy, "will not sell tokenToBuy");
+
+        // executeSwap if storage auction already exists
+        DutchAuction storage auction = dutchAuctions[tokenOut][tradeEnd];
+        if (
+            tradeEnd > block.timestamp &&
+            (address(auction.sell) != address(0) || address(auction.buy) != address(0))
```
pmckelvy1, line 150: do we even need `address(auction.buy) != address(0)`? don't we just need to make sure that the sell token is correct?

contracts/p1/RevenueTrader.sol
```diff
@@ -89,14 +119,121 @@ contract RevenueTraderP1 is TradingP1, IRevenueTrader {
...
+        ) {
+            return executeSwap(auction, amountOut);
+        }
+
+        // === Checks/Effects ===
+
+        require(
+            tradeEnd + dutchAuctionLength > block.timestamp &&
+                block.timestamp + dutchAuctionLength > tradeEnd,
```
pmckelvy1, line 159: i think we want >= here

pmckelvy1: edge case: using `>` makes it so that multiple auctions cannot be swapped in the same block. the first would bump `tradeEnd` (below), and the next would fail here

contracts/libraries/DutchAuctionLib.sol
```diff
@@ -0,0 +1,198 @@
...
+    }
+
+    // === Private ===
+
+    uint192 private constant FIFTEEN_PERCENT = 15e16; // {1}
+    uint192 private constant FIFTY_PERCENT = 50e16; // {1}
+    uint192 private constant EIGHTY_FIVE_PERCENT = 85e16; // {1}
+
+    /// Price Curve:
+    ///   - 1.5 * middlePrice down to the middlePrice for first 20% of auction
```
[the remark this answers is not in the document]
tbrent: Yup, stale comment after I updated it to 15%. Thx for catch

contracts/p1/BackingManager.sol
```diff
@@ -18,8 +19,9 @@ import "./mixins/RecollateralizationLib.sol";
 
 /// @custom:oz-upgrades-unsafe-allow external-library-linking
 contract BackingManagerP1 is TradingP1, IBackingManager {
+    using DutchAuctionLib for DutchAuction;
     using FixLib for uint192;
-    using SafeERC20Upgradeable for IERC20Upgradeable;
```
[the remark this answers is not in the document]
tbrent: I've just replaced it with the non-upgradeable version, is all. There's a `using SafeERC20 for IERC20;` there now

contracts/p1/mixins/Trading.sol
tbrent: Yeah.....good point.

tbrent: Will think on this. The other possible direction is to rename `tradesOpen` to something more narrow, like `nonAtomicTradesOpen` or something

contracts/p1/BackingManager.sol
tbrent: Good suggestion. Looks like this doesn't overflow contract size, either.

contracts/p1/BackingManager.sol
tbrent: I could go either way here. I thought the mismatch error message might not be super clear for a searcher. The BackingManager is highly opinionated about what swap it offers, and all it takes is for some minute appreciation to work out differently for the currently offered swap to change. My thinking was that this kind of error message was more clear, but I don't feel strongly.

contracts/p1/RevenueTrader.sol
tbrent: In P1 I'm going to have it just do `address(auction.sell) != address(0)` since it will be an external call to compare `auction.sell.erc20()` against tokenOut, but in P0 I'll do it the other way.

contracts/p1/BackingManager.sol
tbrent: Yeah...good question, it was something I thought about. 

I think I have a very small preference for returning an empty swap, but the implementation trouble I ran into is that this made it harder to re-use logic or checks that otherwise requires in the `swap()` flow. 

If we weren't running up against BackingManager contract size constraints I would say it's worth it, but as-is it would require breaking up `requireCanTrade` more than I would like to.

PR #779, opened 2023-04-14
Add Collateral plugin for BendDAO supplied ETH

bizzyvinci:
bendWETH which is the token minted when ETH is supplied to BendDAO lending protocol behave in the same way as AAVE AToken. Infact, it can be seen as an AAVE Self Referential token.

Because of some subtle difference, AAVE plugin had to be forked rather than inherited. The differences are
* AAVE IAaveIncentivesController is very different from BendDAO IIncentivesController including the claimReward function and claimer role
* BTokens do not have permit function like AAVE AToken. That's why permit, metaDeposit, and metaWithdraw was removed from StaticBTokenLM

Known issues
* `returns the correct bal (18 decimals)` is failing cause the required [dynamicAmount](https://github.com/bizzyvinci/reserve-protocol/blob/99b7c22c32d3473f872f14409a42f24067143013/test/plugins/individual-collateral/benddao/helpers.ts#L52) to deposit increases a little with time. Everything has to occur at the same time
* Linting is very unhappy. I had to hack my way by setting `max-line-length` to 200


tbrent:
Hey @bizzyvinci 

Thanks for submitting this. I've opened #838 with your commit history as the base. I'm going to close this PR out since we won't be needing it, but you'll remain a contributor on the repo.

PR #782, opened 2023-04-17
Feature/remove delegate call

pmckelvy1:


contracts/p1/mixins/RewardableLib.sol
```diff
@@ -16,32 +16,33 @@ library RewardableLibP1 {
...
-            address(registry.assets[i]).functionDelegateCall(
-                abi.encodeWithSignature("claimRewards()"),
-                "rewards claim failed"
-            );
+        for (uint256 i = 0; i < registry.erc20s.length; ++i) {
+            // empty try/catch because not every erc20 will be wrapped & have a claimRewards func
+            // solhint-disable-next-line
+            try IRewardable(address(registry.erc20s[i])).claimRewards() {} catch {}
         }
+        // weird solidity compiler bug
```
tbrent, line 34: wtfff

contracts/plugins/assets/aave/ATokenFiatCollateral.sol
```diff
@@ -54,9 +54,7 @@ contract ATokenFiatCollateral is AppreciatingFiatCollateral {
     /// Claim rewards earned by holding a balance of the ERC20 token
     /// @dev Use delegatecall
```
tbrent, line 55: Should remove `@dev Use delegatecall` everywhere

contracts/vendor/solmate/ERC4626Rewardable.sol
```diff
@@ -0,0 +1,237 @@
...
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import "./IERC4626.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+
+/**
+ * @dev Implementation of the ERC4626 "Tokenized Vault Standard" as defined in
+ * https://eips.ethereum.org/EIPS/eip-4626[EIP-4626].
+ * copied from openzepplein-contracts/ERC4626.sol, commit hash 91df66c
+ * augmented with rewardability
```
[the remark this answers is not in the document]
tbrent: Isn't it missing an integral on share balances over time?

contracts/plugins/assets/vaults/RewardableERC20Vault.sol
```diff
@@ -0,0 +1,84 @@
...
+
+    function _claimAccountRewards(address account) internal {
+        uint256 claimableRewards = accumulatedRewards[account] - claimedRewards[account];
+        emit RewardsClaimed(IERC20(address(rewardToken)), claimableRewards);
+        if (claimableRewards == 0) return;
+        claimedRewards[account] = accumulatedRewards[account];
+        rewardToken.safeTransfer(account, claimableRewards);
+    }
+
+    function _beforeWithdraw(
```
tbrent, line 67: Don't we need to do all this logic on each transfer of the share, as well?

contracts/p1/mixins/RewardableLib.sol
pmckelvy1: actually turned out to be a weird hardhat-ethers bug, thankfully. reproduced and submitted here: https://github.com/NomicFoundation/hardhat/issues/3880

ill fix this comment

contracts/plugins/assets/vaults/RewardableERC20Vault.sol
```diff
@@ -0,0 +1,84 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+pragma solidity ^0.8.17;
+
+import "../../../interfaces/IRewardable.sol";
+import "../../../vendor/solmate/ERC4626Rewardable.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
+import "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+abstract contract RewardableERC20Vault is IRewardable, ERC4626Rewardable {
```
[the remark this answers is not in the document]
pmckelvy1: scratch

contracts/vendor/solmate/ERC4626Rewardable.sol
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    /** @dev See {IERC4626-previewMint}. */
+    function previewMint(uint256 shares) public view virtual override returns (uint256) {
+        return _convertToAssets(shares, Math.Rounding.Up);
+    }
+
+    /** @dev See {IERC4626-previewWithdraw}. */
+    function previewWithdraw(uint256 assets) public view virtual override returns (uint256) {
+        return _convertToShares(assets, Math.Rounding.Up);
```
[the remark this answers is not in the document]
pmckelvy1: it's always rounding in favor of the vault. in this specific function, the reasoning is "the number of assets wanted by the withdrawer should cost the greatest amount of shares to obtain"

test/scenario/MaxBasketSize.test.ts
```diff
@@ -574,20 +580,21 @@ describe(`Max Basket Size - P${IMPLEMENTATION}`, () => {
...
-          },
-          {
-            contract: backingManager,
-            name: 'RewardsClaimed',
-            args: [aaveToken.address, rewardAmount],
-            emitted: true,
-          },
-        ])
+        await backingManager.claimRewards()
+        // await expectEvents(backingManager.claimRewards(), [
```
tbrent, line 584: I think these can be uncommented, and the contract changed to be the 4626 ERC20s

test/scenario/cETH.test.ts
```diff
@@ -120,8 +123,8 @@ describe(`CToken of self-referential collateral (eg cETH) - P${IMPLEMENTATION}`,
 
     // cETH
     cETH = await (
-      await ethers.getContractFactory('CTokenMock')
-    ).deploy('cETH Token', 'cETH', weth.address)
+      await ethers.getContractFactory('CTokenVaultMock')
+    ).deploy('cWBTC Token', 'cWBTC', weth.address, compToken.address, compoundMock.address)
```
tbrent, line 127: nit: cETH

tbrent:
LGTM

test/scenario/MaxBasketSize.test.ts
pmckelvy1: so if we really want to test that the events get fired, we would want to iterate over all the tokens in the basket and do some logic to know whether to expect aave or comp as a reward. we would also need to use anyValue because only one of each rewarding token will actually end up firing an event with any value but 0.

given that we test these events in dozens of other tests, and that this test is really just a gas test, i think we can just call `claimRewards` without using `expectEvents`

pmckelvy1:
i think its worth adding VERSION fields to the collateral plugins

tbrent:
> i think its worth adding VERSION fields to the collateral plugins

That makes sense, agree. I think we should create a new mixin specifically for collateral plugins that mirrors `contracts/mixin/Versioned.sol`. For now we can keep the version strings the same, and then after the monorepo refactor we'll allow them to diverge.

PR #783, opened 2023-04-18
Feat stake and delegate

julianmrodri:
* Allow to `stakeAndDelegate` in single transaction

### DIFF
--- a/contracts/p1/StRSRVotes.sol
+++ b/contracts/p1/StRSRVotes.sol
@@ -134,12 +134,17 @@ contract StRSRP1Votes is StRSRP1, IStRSRVotes {
         _delegate(signer, delegatee);
     }
 
-    /// Stakes an RSR `amount` on the corresponding RToken and delegates
+    /// Stakes an RSR `amount` on the corresponding RToken and allows to delegate
     /// votes from the sender to `delegatee` or self
-    function stakeAndDelegate(uint256 rsrAmount, address delegatee) public {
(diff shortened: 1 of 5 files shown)


contracts/p1/StRSRVotes.sol
```diff
@@ -134,6 +134,19 @@ contract StRSRP1Votes is StRSRP1, IStRSRVotes {
...
     }
 
+    /// Stakes an RSR `amount` on the corresponding RToken and allows to delegate
+    /// votes from the sender to `delegatee` or self
+    function stakeAndDelegate(uint256 rsrAmount, address delegatee) external {
+        stake(rsrAmount);
+        address msgSender = _msgSender();
+
+        // Delegate if a specific delegatee is provided or if no delegate yet
+        if (delegatee != address(0) || delegates(msgSender) == address(0)) {
```
tbrent, line 144: No need to re-delegate if the delegate has already been set, so I'm thinking the condition could be `delegatee != address(0) && delegates(msgSender) != delegatee`. Should capture the case that is here and then some.

tbrent: Hmmm it doesn't capture the delegate-to-self via zero_address, though, does it...

tbrent: Nvm, LGTM

julianmrodri: Implemented the change discussed offline.

tbrent:
LGTM! Can merge once tests finish.

PR #786, opened 2023-04-21
Refactoring: merge FiatCollateral, NonFiatCollateral, EURFiatCollateral and SelfReferentialCollateral

apehex:
Hey there! Here's a little experiment I did while working on collateral plugins.

I propose to merge `FiatCollateral.sol`, `NonFiatCollateral.sol`, `EURFiatCollateral.sol` and `SelfReferentialCollateral.sol` into a single contract.

It would help reduce redundancy in collateral plugin contracts and further simplify the creation of plugins.


apehex:
## Price Calculation

### Current Logic

In the formula:

$$\begin{align}
\{UoA\/tok\} = \{UoA\/target\} * \{target\/ref\} * \{ref\/tok\}
\end{align}$$

Only one factor is addressed by the base class `Asset.sol`, with the property `chainlinkFeed`.

It is up to the plugin writers to:

- decide the units and assign the `chainlinkFeed` to one factor
- eventually add another oracle
- and essentially rewrite the `tryPrice` method

### Proposition

Instead, we could use:

$$\begin{align}
\{UoA\/tok\} = \{UoA\/target\} * \frac{\{UoA\/ref\}}{\{UoA\/target\}} * \{ref\/tok\}
\end{align}$$

And specifically assign:

- one oracle to $\{UoA\/target\}$, `uoaPerTargetOracle`
- and another to $\{UoA\/ref\}$, `uoaPerRefOracle`

As explained below, if two units are the same then the corresponding oracle has address `0x0` and the code replaces the corresponding factor by one.

## Implementation

Now there's a single contract for all cases: [`Collateral.sol`][merged-contract].
It defines `tryPrice` once and adds `uoaPerTargetOracle` and `uoaPerTargetOracleTimeout` to the `CollateralConfig`.

Null addresses are allowed for both oracles: it means the corresponding factor is 1.
For example, if $\{UoA\} = \{USD\}$ and $\{target\} = \{USD\}$, the collateral does not need a $\{UoA\/target\}$ oracle.
Then `config.uoaPerTargetOracle = address(0)`.

## Pros

- less code:
  - the function `tryPrice` is written once for all plugins (except Convex)
  - easier to understand / extend, especially for new coders
- more explicit:
  - the units for each oracle are known
- more practical:
  - all the oracles use USD as reference, which will always be available
  - on the contrary, there may not be a $\{target\/ref\}$ oracle

## Cons

Essentially it requires to redo the tests for each plugin.

While the changes are minimal (see below), the codebase for the tests is expensive and they'd need to be reviewed.

Since this doesn't add any functionality, the workload is likely a deal-breaker.

## Migrating The Existing Collaterals

Instead of writing new contracts for the collateral variants (non-fiat etc), the new design just requires the correct oracle addresses.

`uoaPerRefOracle` is used for the peg $\{target\/ref\}$.
The two oracles cannot be switched, even when a factor 1 is in the above formula.

You can adapt the examples below to the specific unit of a given plugin.

### Fiat Collaterals

```solidity
{
    ...,
    uoaPerTargetOracle: address(0), // {UoA} = {target}, IE {UoA/target} = 1
    uoaPerRefOracle: networkConfig[chainId].chainlinkFeeds.DAI // {USD/DAI} for example
    ...,
}
```

### Non-fiat Collaterals

```solidity
{
    ...,
    uoaPerTargetOracle: '0xf4030086522a5beea4988f8ca5b36dbc97bee88c, // {USD/BTC}
    uoaPerRefOracle: '0x639545836d8b177054cefafe6942efe798ce6575' // {USD/wBTC} and the token would be cwBTC for example
    ...,
}
```

### Self-referential Collaterals

```solidity
{
    ...,
    uoaPerTargetOracle: networkConfig[chainId].chainlinkFeeds.ETH, // {USD/ETH}
    uoaPerRefOracle: address(0) // with {target/ref} = {ETH/wETH} = 1
    ...,
}
```

### EUR-fiat Collaterals

```solidity
{
    ...,
    uoaPerTargetOracle: '0xb49f677943bc038e9857d61e7d053caa2c1734c1', // {USD/EUR}
    uoaPerRefOracle: '' // I don't have any example yet, but any feed for a stable Euro coin
    ...,
}
```

[merged-contract]: https://github.com/apehex/protocol/blob/draft/factor-collateral-plugins/contracts/plugins/assets/Collateral.sol

tbrent:
Hey, this is awesome! Love to see this kind of thing (also: great job writing in our style)

I definitely agree on the cost/benefit analysis, due to the tests. I'm glad you didn't just go for it, I wouldn't wish that suffering on anyone...

Thoughts:
- I think we can go all the way and absorb AppreciatingFiatCollateral too. The refresh() function differs but it can be unified using the same methodology.
- If we go this route I'd like to see the oracle storage variables made immutable, like [this](https://github.com/reserve-protocol/protocol/blob/d75348dc28fb73f8e7610d95fa86fcd39b5b0434/contracts/plugins/assets/convex/PoolTokens.sol#L65), so they just become part of the bytecode. That would require pulling them out of any structs, so it's a larger refactor
- There would still be some mild amount of unnecessary gas costs due to == 0 checks, but it would be worth it

Thanks for starting this!

apehex:
Damn that's (honestly) unexpected!!
I had lost all faith in this idea after writing it :joy: 

Great thought too!
That's exactly the way I see it: for this PR to be useful, we'd have to combine it with any pending improvement you have (?).

The pain of testing will be real!...
It would have been super fishy if I went ahead with those, like randomly offering to carry people's stuff in the street "yeah thanks no thanks, I can take care of my things tyyy".

PR #787, opened 2023-04-24
Feat/upgrades - Upgradeability checks

julianmrodri:
* Adds new tasks and modifies deployment script to be able to use the Hardhat OZ Upgrade Plugin
* `forceImport`: Allows to import details and storage layout of already deployed implementations  (only on Mainnet)
* `validateUpgrade`:  Checks the current code if its upgrade safe vs a previously deployed version
* Modifies the `deploy_implementations` step in deployment scripts, to ensure we use `deployImplementation` or `prepareUpgrade` which will validate if they are upgrade safe and also track the new implementations in the network file.
* Adds the `.openzeppelin` details for version `2.0.0` and `2.1.0` (these are required for future upgrades). Only for Mainnet.
* Bumps HH and OZ dependencies to use the features we need.


tasks/upgrades/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+) => {
+  // Check trading lib defined
+  if (!deployments.tradingLib) {
+    throw new Error(
+      `Missing deployed TradingLib for version ${version} in network ${hre.network.name}`
+    )
+  }
+
+  // Check implementations are defined
+  if (
```
pmckelvy1, line 17: might be worth splitting these into their own error statements so that we know exactly which ones are missing?

pmckelvy1 (changes requested):

tasks/upgrades/utils.ts
julianmrodri: I can do the change. I dont think its a big deal to leave it like it is (its just taking a looking at the deployment file and check whats wrong)  but but there is no harm in being more specific. I'll do this.

scripts/deployment/phase1-common/2_deploy_implementations.ts
```diff
@@ -113,98 +168,242 @@ async function main() {
...
   const RevTraderImplFactory = await ethers.getContractFactory('RevenueTraderP1')
-  revTraderImpl = <RevenueTraderP1>await RevTraderImplFactory.connect(burner).deploy()
-  await revTraderImpl.deployed()
+  let revTraderImplAddr = ''
+  if (!upgrade) {
+    revTraderImplAddr = await upgrades.deployImplementation(RevTraderImplFactory, {
+      kind: 'uups',
+    })
+  } else {
+    // We can use any of the two addresses (should be equal)
```
pmckelvy1, line 321: *should*, but might not be. probably worth at least verifying that they are equal? or making sure we verify against both?

tbrent:
@julianmrodri should this one be closed?

PR #790, opened 2023-04-25
Issuance when trading paused

tbrent:
Issuance was reverting if trading was paused. Fixed


julianmrodri:
image

There are failures on the test. Do you want to take care of that? If not let me know.

julianmrodri (commented):
LGTM but we should fix the tests.

PR #791, opened 2023-04-26
Itrade dutch auction

### LINKED ISSUE #796: Melt() before trading
also assetRegistry.refresh() before revenue auctions

### LINKED ISSUE #798: Dutch Auctions
https://www.notion.so/myreserve/Atomic-Trading-Design-Doc-32f53e6762f24ef1a635a9b68e959f72

### LINKED ISSUE #802: Keep backingBuffer in BackingManager to pay for slippage during rebalances
Image

The `needed = needed.mul(FIX_ONE.plus(backingBuffer));` line needs to be above the RSR minting block if we want to _keep_ capital around up to the BackingBuffer.

Or, alternatively, we can keep it where it is but not forward the RToken in the loop at the bottom up to the correct amount.

tbrent:
Resolves #796 
Resolves #802 
Resolves #798

### DIFF
--- a/contracts/p1/Broker.sol
+++ b/contracts/p1/Broker.sol
@@ -23,6 +23,8 @@ contract BrokerP1 is ComponentP1, IBroker {
     using Clones for address;
 
     uint48 public constant MAX_AUCTION_LENGTH = 604800; // {s} max valid duration - 1 week
+    uint48 public constant MIN_AUCTION_LENGTH = ONE_BLOCK; // {s} min auction length - 1 block
+    // warning: blocktime <= 12s assumption
 
     IBackingManager private backingManager;
     IRevenueTrader private rsrTrader;
@@ -148,7 +150,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     /// @custom:governance
     function setBatchAuctionLength(uint48 newAuctionLength) public governance {
         require(
-            newAuctionLength > 0 && newAuctionLength <= MAX_AUCTION_LENGTH,
+            newAuctionLength >= MIN_AUCTION_LENGTH && newAuctionLength <= MAX_AUCTION_LENGTH,
             "invalid batchAuctionLength"
         );
         emit BatchAuctionLengthSet(batchAuctionLength, newAuctionLength);
@@ -169,7 +171,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     /// @custom:governance
     function setDutchAuctionLength(uint48 newAuctionLength) public governance {
         require(
-            newAuctionLength > 0 && newAuctionLength <= MAX_AUCTION_LENGTH,
+            newAuctionLength >= MIN_AUCTION_LENGTH && newAuctionLength <= MAX_AUCTION_LENGTH,
             "invalid dutchAuctionLength"
         );
         emit DutchAuctionLengthSet(dutchAuctionLength, newAuctionLength);
@@ -204,7 +206,6 @@ contract BrokerP1 is ComponentP1, IBroker {
             address(trade),
             req.sellAmount
         );
-
         trade.init(this, caller, gnosis, batchAuctionLength, req);
         return trade;
     }
--- a/contracts/p1/Broker.sol
+++ b/contracts/p1/Broker.sol
@@ -141,7 +141,7 @@ contract BrokerP1 is ComponentP1, IBroker {
             "invalid batchTradeImplementation address"
         );
 
-        emit TradeImplementationSet(batchTradeImplementation, newTradeImplementation);
+        emit BatchTradeImplementationSet(batchTradeImplementation, newTradeImplementation);
         batchTradeImplementation = newTradeImplementation;
     }
 
@@ -151,7 +151,7 @@ contract BrokerP1 is ComponentP1, IBroker {
             newAuctionLength > 0 && newAuctionLength <= MAX_AUCTION_LENGTH,
             "invalid batchAuctionLength"
         );
-        emit AuctionLengthSet(batchAuctionLength, newAuctionLength);
+        emit BatchAuctionLengthSet(batchAuctionLength, newAuctionLength);
         batchAuctionLength = newAuctionLength;
     }
 
@@ -162,7 +162,7 @@ contract BrokerP1 is ComponentP1, IBroker {
             "invalid dutchTradeImplementation address"
         );
 
-        emit TradeImplementationSet(dutchTradeImplementation, newTradeImplementation);
+        emit DutchTradeImplementationSet(dutchTradeImplementation, newTradeImplementation);
         dutchTradeImplementation = newTradeImplementation;
     }
 
@@ -172,7 +172,7 @@ contract BrokerP1 is ComponentP1, IBroker {
             newAuctionLength > 0 && newAuctionLength <= MAX_AUCTION_LENGTH,
             "invalid dutchAuctionLength"
         );
-        emit AuctionLengthSet(dutchAuctionLength, newAuctionLength);
+        emit DutchAuctionLengthSet(dutchAuctionLength, newAuctionLength);
         dutchAuctionLength = newAuctionLength;
     }
 
--- a/contracts/p1/Broker.sol
+++ b/contracts/p1/Broker.sol
@@ -31,7 +31,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     // The Batch Auction Trade contract to clone on openTrade(). Governance parameter.
     ITrade public batchTradeImplementation;
 
-    // The Gnosis contract to init each trade with. Governance parameter.
+    // The Gnosis contract to init batch auction trades with. Governance parameter.
     IGnosis public gnosis;
 
     // {s} the length of a Gnosis EasyAuction. Governance parameter.
@@ -183,7 +183,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     // === Private ===
 
     function newBatchAuction(TradeRequest memory req, address caller) private returns (ITrade) {
-        // In the future we'll have more sophisticated choice logic here, probably by trade size
+        require(batchAuctionLength > 0, "batchAuctionLength unset");
         GnosisTrade trade = GnosisTrade(address(batchTradeImplementation).clone());
         trades[address(trade)] = true;
 
@@ -208,6 +208,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     }
 
     function newDutchAuction(TradeRequest memory req, ITrading caller) private returns (ITrade) {
+        require(dutchAuctionLength > 0, "dutchAuctionLength unset");
         DutchTrade trade = DutchTrade(address(dutchTradeImplementation).clone());
         trades[address(trade)] = true;
 
--- a/contracts/p1/Broker.sol
+++ b/contracts/p1/Broker.sol
@@ -138,7 +138,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     function setBatchTradeImplementation(ITrade newTradeImplementation) public governance {
         require(
             address(newTradeImplementation) != address(0),
-            "invalid Trade Implementation address"
+            "invalid batchTradeImplementation address"
         );
 
         emit TradeImplementationSet(batchTradeImplementation, newTradeImplementation);
@@ -159,7 +159,7 @@ contract BrokerP1 is ComponentP1, IBroker {
     function setDutchTradeImplementation(ITrade newTradeImplementation) public governance {
         require(
             address(newTradeImplementation) != address(0),
-            "invalid Trade Implementation address"
+            "invalid dutchTradeImplementation address"
         );
 
         emit TradeImplementationSet(dutchTradeImplementation, newTradeImplementation);
--- a/contracts/p1/Broker.sol
+++ b/contracts/p1/Broker.sol
@@ -9,6 +9,7 @@ import "../interfaces/IMain.sol";
 import "../interfaces/ITrade.sol";
 import "../libraries/Fixed.sol";
 import "./mixins/Component.sol";
+import "../plugins/trading/DutchTrade.sol";
 import "../plugins/trading/GnosisTrade.sol";
 
 // Gnosis: uint96 ~= 7e28
@@ -27,14 +28,14 @@ contract BrokerP1 is ComponentP1, IBroker {
     IRevenueTrader private rsrTrader;
     IRevenueTrader private rTokenTrader;
 
-    // The trade contract to clone on openTrade(). Governance parameter.
-    ITrade public tradeImplementation;
+    // The Batch Auction Trade contract to clone on openTrade(). Governance parameter.
+    ITrade public batchTradeImplementation;
 
     // The Gnosis contract to init each trade with. Governance parameter.
     IGnosis public gnosis;
 
-    // {s} the length of an auction. Governance parameter.
-    uint48 public auctionLength;
+    // {s} the length of a Gnosis EasyAuction. Governance parameter.
+    uint48 public batchAuctionLength;
 
     // Whether trading is disabled.
     // Initially false. Settable by OWNER. A trade clone can set it to true via reportViolation()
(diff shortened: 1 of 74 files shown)


contracts/p1/mixins/Trading.sol
```diff
@@ -108,15 +120,27 @@ abstract contract TradingP1 is Multicall, ComponentP1, ReentrancyGuardUpgradeabl
...
         IERC20Upgradeable(address(sell)).safeApprove(address(broker), 0);
         IERC20Upgradeable(address(sell)).safeApprove(address(broker), req.sellAmount);
-        ITrade trade = broker.openTrade(req);
 
+        // Only start the next auction back-to-back if msgSender is self
+        if (_msgSender() != address(this)) {
+            // Require at least 1 empty block between auctions of the same kind
+            // This gives space for someone to start one of the opposite kinds of auctions
+            if (kind == TradeKind.DUTCH_AUCTION) {
+                require(block.number > lastSettlement[TradeKind.DUTCH_AUCTION] + 1, "wait 1 block");
```
pmckelvy1: i thought we wanted to wait 1 full length of an auction?

pmckelvy1 (changes requested):

contracts/p1/Broker.sol
```diff
@@ -158,10 +180,53 @@ contract BrokerP1 is ComponentP1, IBroker {
...
+        }
+
+        // == Interactions ==
+        IERC20Upgradeable(address(req.sell.erc20())).safeTransferFrom(
+            caller,
+            address(trade),
+            req.sellAmount
+        );
+
+        trade.init(this, caller, gnosis, batchAuctionLength, req);
```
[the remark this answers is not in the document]
tbrent: yes, the `req.minBuyAmount` would go unused, so this seemed more clear. the `init()` function isn't on the `ITrade` interface anyway

contracts/p1/Broker.sol
```diff
@@ -158,10 +180,53 @@ contract BrokerP1 is ComponentP1, IBroker {
...
+        return trade;
+    }
+
     /**
      * @dev This empty reserved space is put in place to allow future versions to add new
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[44] private __gap;
+    uint256[43] private __gap;
```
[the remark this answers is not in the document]
tbrent: They fit in the same slot; an address (even if typed) is 20 bytes, and uint48 is 6 bytes

contracts/p1/RevenueTrader.sol
```diff
@@ -89,10 +99,25 @@ contract RevenueTraderP1 is TradingP1, IRevenueTrader {
...
+
+    function distributeRevenue() private {
+        uint256 bal = tokenToBuy.balanceOf(address(this));
+        tokenToBuy.safeApprove(address(distributor), 0);
+        tokenToBuy.safeApprove(address(distributor), bal);
+        distributor.distribute(tokenToBuy, bal);
+    }
+
+    // // {UoA}
+    // uint192 maxTradeVolume = fixMin(sell_.maxTradeVolume(), buy_.maxTradeVolume());
```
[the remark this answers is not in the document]
tbrent: HA no that's a WIP for when I split up `manageToken()`, I'll mark it.

contracts/p1/mixins/Trading.sol
```diff
@@ -108,15 +120,27 @@ abstract contract TradingP1 is Multicall, ComponentP1, ReentrancyGuardUpgradeabl
     // This is reentrancy-safe because we're using the `nonReentrant` modifier on every method of
     // this contract that changes state this function refers to.
     // slither-disable-next-line reentrancy-vulnerabilities-1
-    function tryTrade(TradeRequest memory req) internal nonReentrant {
+    function tryTrade(TradeRequest memory req, TradeKind kind) internal nonReentrant {
```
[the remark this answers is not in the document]
tbrent: It's orthogonal: `RecollateralizationLib.sol` prepares a trade without defining what TradeKind should be used to execute it. 

Now, the `minBuyAmount` isn't used in the case of TradeKind.DUTCH_AUCTION, so it's possible there's something that should change here.

contracts/p1/mixins/Trading.sol
tbrent: Realized it was unnecessary. Let me know if you think it is necessary. It would mean slower rebalancing.

contracts/p1/mixins/Trading.sol
```diff
@@ -108,14 +120,38 @@ abstract contract TradingP1 is Multicall, ComponentP1, ReentrancyGuardUpgradeabl
...
         IERC20Upgradeable(address(sell)).safeApprove(address(broker), req.sellAmount);
-        ITrade trade = broker.openTrade(req);
+
+        // Only allow starting the next auction back-to-back if msgSender is self
+        // Only time this happens is BackingManager.settleTrade() -> BackingManager.rebalance()
+        // TODO is there a better way to do this?
+        if (_msgSender() != address(this)) {
+            // Warning, Assumption: blocktime <= 12s
+            // Require at least 1 block between auctions of the same kind
+            // This gives space for someone to start one of the opposite kinds of auctions
```
tbrent, line 137: We can't compare against a value saved during `settleTrade()` because we want to support the usecase where settleTrade() is higher up in the block. `endTime()` is the best thing we have laying around. We don't know the block number the trade will end at ahead of time.

contracts/p1/mixins/Trading.sol
pmckelvy1: i guess it's really only necessary during an awkward transition phase where this is still semi-manual but bots are coming online...probably not necessary to build specifically for that temporary case.

contracts/p1/RevenueTrader.sol
```diff
@@ -48,24 +71,34 @@ contract RevenueTraderP1 is TradingP1, IRevenueTrader {
...
+                assetRegistry.refresh();
+                furnace.melt();
+            } else {
+                // otherwise, refresh just buy + sell
+                sell.refresh();
+                buy.refresh();
+            }
+        }
+
+        require(address(trades[erc20]) == address(0), "trade open");
```
akshatmittal, line 99: Any reason to why they were moved here? Ideally these would be the first checks in order to fail fast. These are reverts anyway, so previous progress is always lost.

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -0,0 +1,220 @@
...
+    function canSettle() external view returns (bool) {
+        return status == TradeStatus.OPEN && (bidder != address(0) || block.timestamp >= endTime);
+    }
+
+    // === Bid Helper ===
+
+    /// Calculates how much buy token is needed to purchase the lot, at a particular timestamp
+    /// Price Curve:
+    ///   - 1.5 * middlePrice down to the middlePrice for first 15% of auction
+    ///   - middlePrice down to lowPrice for the last 80% of auction
```
akshatmittal, line 191: Comment, should be `85%`

contracts/p1/RevenueTrader.sol
```diff
@@ -88,15 +121,23 @@ contract RevenueTraderP1 is TradingP1, IRevenueTrader {
             maxTradeSlippage
         );
 
-        if (launch) {
-            tryTrade(req);
-        }
+        require(launch, "trade not worth launching");
```
akshatmittal, line 124: Since we only check if the trade is actually worth launching and not which trade it is, would frontrunning be an issue?

(Someone could frontrun a batch auction with a dutch auction and vice-versa, maybe a DoS vector or malicious intent)

test/utils/trades.ts
```diff
@@ -73,3 +73,56 @@ export const toMinBuyAmt = (
...
+
+  const price = progression.lt(fp('0.15'))
+    ? highPrice.sub(highPrice.sub(middlePrice).mul(progression).div(fp('0.15')))
+    : middlePrice.sub(
+        middlePrice
+          .sub(lowPrice)
+          .mul(progression.sub(fp('0.15')))
+          .div(fp('0.85'))
+      )
+  // console.log(
```
pmckelvy1, line 118: delete

test/utils/trades.ts
```diff
@@ -73,3 +73,56 @@ export const toMinBuyAmt = (
...
+  const auctionVolume = outAmount.mul(sellHigh).div(fp('1'))
+  const slippage = maxTradeSlippage
+    .mul(
+      fp('1').sub(
+        auctionVolume.sub(minTradeVolume).mul(fp('1')).div(maxTradeVolume.sub(minTradeVolume))
+      )
+    )
+    .div(fp('1'))
+
+  // console.log('slippage: ', slippage)
```
pmckelvy1, line 104: delete

test/Recollateralization.test.ts
```diff
@@ -3088,6 +3113,141 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
...
+        it('Should bid in final second of auction and launch another auction', async () => {
+          await backingManager.rebalance(TradeKind.DUTCH_AUCTION)
+          const trade = await ethers.getContractAt(
+            'DutchTrade',
+            await backingManager.trades(token0.address)
+          )
+          await token1.connect(addr1).approve(trade.address, initialBal)
+
+          // Snipe auction at 1s left
+          await advanceToTimestamp((await getLatestBlockTimestamp()) + auctionLength - 3)
```
pmckelvy1, line 3224: curious, why not - 2?

test/Recollateralization.test.ts
```diff
@@ -3088,6 +3113,141 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
...
+            ),
+            bn('1e12') // decimals
+          )
+          expect(await backingManager.tradesOpen()).to.equal(1)
+          expect(await token1.balanceOf(backingManager.address)).to.equal(expected)
+
+          // Should launch another auction: RSR -> token1
+          expect(await backingManager.trades(token0.address)).to.equal(ZERO_ADDRESS)
+          expect(await backingManager.trades(rsr.address)).to.not.equal(ZERO_ADDRESS)
+          expect(await backingManager.tradesOpen()).to.equal(1)
```
pmckelvy1, line 3248: already checked above

test/Recollateralization.test.ts
```diff
@@ -3088,6 +3113,141 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
...
+        })
+      })
```
pmckelvy1, line 3250: probably worth testing a couple more scenarios:
1) token dutch auction, rsr dutch auction, complete auction and check full recollateralization
2) token dutch auction, timeout rsr dutch auction and send it to a batch auction, complete and check full recollateralization

honestly you could probably just extend the above test to cover (1)

test/Recollateralization.test.ts
```diff
@@ -3088,6 +3113,141 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
...
+          await expect(backingManager.rebalance(TradeKind.DUTCH_AUCTION)).to.emit(
+            backingManager,
+            'TradeStarted'
+          )
+        })
+
+        it('Should only run 1 trade at a time', async () => {
+          await backingManager.rebalance(TradeKind.DUTCH_AUCTION)
+          await expect(backingManager.rebalance(TradeKind.DUTCH_AUCTION)).to.be.revertedWith(
+            'already rebalancing'
```
pmckelvy1, line 3152: we should add another scenario that explicitly checks for `already rebalancing` in the "buffer" block

pmckelvy1: ie. run dutch auction => jump to tradeEnd[DUTCH_AUCTION] + ONE_BLOCK => try to rerun dutch auction and check reverted with "already rebalancing"

pmckelvy1 (changes requested):

contracts/p1/RevenueTrader.sol
tbrent: Agree, ideally they'd be first. This is an instance though of our larger RCEI pattern, where it's important to do the checks _after_ the first set of refresh interactions, since they could plausibly change the outcome of the checks. 

This wasn't labeled well as trying to follow RCEI, so I've added that to the `@custom:interaction` comment at the top of the function. The `tokenToBuy` is immutable after init and the assetRegistry assets can only be changed under governance, which is why it's acceptable to place these lines above the refresh() interactions.

contracts/p1/RevenueTrader.sol
tbrent: Right. So in the BackingManager we have DoS protection from the `tradeEnd[kind] < block.timestamp + ONE_BLOCK` condition, and I think that's important. This requires there to be at least one block between auctions of the same kind. Otherwise plausibly you could run into a situation where an auction kind is not clearing, and an attacker burns ETH for a while in order to prevent recollateralization. 

In the RevenueTrader it's more complicated because the RevenueTrader _does_ run parallel auctions as long as they are for different ERC20s. We could do the same kind of DoS protection, but only if we split this out by ERC20. I think I don't see the value here, though. With the BackingManager there is some urgency to doing the trading in a timely manner, but the same isn't true of the RevenueTrader. The DoS attacker basically has to burn ETH forever, and it's not clear how they would benefit from this. 

I might be missing something, but I don't see why it would be valuable for the RevenueTrader to do the same kind of DoS protection.

scripts/deployment/phase3-rtoken/rTokenConfig.ts
```diff
@@ -21,7 +21,7 @@ export const rTokenConfig: { [key: string]: IRToken } = {
         rewardRatio: bn('1069671574938'), // approx. half life of 90 days
         unstakingDelay: bn('1209600'), // 2 weeks
         tradingDelay: bn('14400'), // (the delay _after_ default has been confirmed) 4 hours
-        auctionLength: bn('900'), // 15 minutes
+        batchAuctionLength: bn('900'), // 15 minutes
```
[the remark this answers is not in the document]
tbrent: Yep, good catch. We don't do phase3 of deployments currently but this should still be updated.

test/Recollateralization.test.ts
```diff
@@ -1314,7 +1331,7 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
         expect(await rToken.totalSupply()).to.be.gt(issueAmount) // New RToken minting
 
         // Check price in USD of the current RToken
-        expect(await rToken.basketsNeeded()).to.equal(await rToken.totalSupply()) // no haircut
+        expect(await rToken.basketsNeeded()).to.be.gte(await rToken.totalSupply()) // no haircut
```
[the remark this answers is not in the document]
tbrent: It's due to the backing buffer (some collateral is set aside, before minting new RToken), but I think for this test it's better to just set the backing buffer to 0.

test/Broker.test.ts
```diff
@@ -393,588 +499,858 @@ describe(`BrokerP${IMPLEMENTATION} contract #fast`, () => {
...
-      await token0.connect(owner).mint(trade.address, amount)
-      await expect(
-        trade.init(
-          broker.address,
-          backingManager.address,
-          gnosis.address,
-          config.auctionLength,
-          tradeRequest
+    context('GnosisTrade', () => {
+      it('Should initialize GnosisTrade correctly - only once', async () => {
```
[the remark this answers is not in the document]
tbrent: I'll pull out the shared stuff as much as possible, but it's not that much.

test/utils/trades.ts
tbrent: really useful for debugging! but okay, yeah, deleting

test/Recollateralization.test.ts
tbrent: `rebalance()` occurs at 0/300; `token1.approve()` occurs at 1/300; it fasts forward 297s to get to 298/300; and then `bid()` executes at 299/300

test/Recollateralization.test.ts
tbrent: it's good you suggested this because the require condition was actually backwards, adding ONE_BLOCK to the wrong side

contracts/p1/RevenueTrader.sol
tbrent: aaaand that should be in a comment shouldn't it

tbrent: Actually, I want to get rid of even that. No reason to settle for reetrancy-safe-unless-governance-action when we can get the real thing.

PR #793, opened 2023-04-27
Roles changes in FacadeWrite and protocol

julianmrodri:
* Only grant `OWNER` role to RToken creator. `PAUSER`, `SHORT_FREEZER`, `LONG_FREEZER` have to be done explicitly. Impacts FacadeWrite, Deployer and Auth.
* Allows pausing multiple `pausers`, `shortFreezers`, and `longFreezers` via FacadeWrite.
* Adapts tests and scripts.

### DIFF
--- a/contracts/mixins/Auth.sol
+++ b/contracts/mixins/Auth.sol
@@ -21,8 +21,7 @@ abstract contract Auth is AccessControlUpgradeable, IAuth {
      *                    and rewards payout
      *  - Issuance Paused: disallow issuance
      *
-     * Typically freezing thaws on its own in a predetemined number of blocks.
-     *   However, OWNER can also freeze forever.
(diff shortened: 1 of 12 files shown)


contracts/mixins/Auth.sol
```diff
@@ -21,8 +21,7 @@ abstract contract Auth is AccessControlUpgradeable, IAuth {
      *                    and rewards payout
      *  - Issuance Paused: disallow issuance
      *
-     * Typically freezing thaws on its own in a predetemined number of blocks.
-     *   However, OWNER can also freeze forever.
```
tbrent, line 25: nit: this is still true, can leave it in

tbrent:
LGTM!

contracts/mixins/Auth.sol
tbrent: I can handle adding it back

PR #808, opened 2023-05-04
add way to start dust auction

jankjr:
Adds a method to manually trigger a trade for any token held by the revenue trader.

It ignores the min trade size restrictions and will allow BATCH_AUCTION's if the oracle is currently broken. Works like a normal call to `RevenueTrader.manageTokens(...)`

If the dust balance is stuck on the `BackingManager` a call to  `forwardRevenue(...)` will forward the dust balances to the `rTokenTrader` and `rsrTrader`, so no changes are needed.


contracts/p1/BackingManager.sol
```diff
@@ -99,6 +99,84 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+
+        // checks: Make sure we don't need to rebalance
+        BasketRange memory basketsHeld = basketHandler.basketsHeldBy(address(this));
+        require(basketsHeld.bottom >= rToken.basketsNeeded(), "undercollateralized");
+
+        // checks: We don't want to sell off basket tokens
+        // @to reviewer: I know that BasketHandletP1 can return just what we need
+        // So i'm not sure if this is the best way to do this otherwise,
+        // so we're quoting for 1 unit using .quote() and just using the returned ERC20s
+        (address[] memory erc20s, ) = basketHandler.quote(FIX_ONE, RoundingMode.CEIL);
```
tbrent, line 135: Simpler to do `basketHandler.quantity(sellToken) == 0`

contracts/p1/BackingManager.sol
```diff
@@ -99,6 +99,84 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+        require(rTokenPriceUSD != 0, "cannot price rToken");
+
+        IAsset sellAsset = assetRegistry.toAsset(sellToken);
+
+        // checks: make sure we even hold a balance
+        uint192 sellQty = sellAsset.bal(address(this));
+        require(sellQty == 0, "zero balance");
+
+        // respect maxTradeVolume
+        uint192 maxTradeSize = sellAsset.maxTradeVolume();
```
tbrent, line 148: The units are different here; `maxTradeVolume()` returns a {UoA}, but `maxTradeSize` is a {tok}. Checkout `TradeLib.maxTradeSize()`. Something else it does different is it also takes into account constraints from both sides of the trade.

contracts/p1/BackingManager.sol
```diff
@@ -99,6 +99,84 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+        sellQty = sellQty > maxTradeSize ? maxTradeSize : sellQty;
+
+        // checks: make sure all tokens can be priced
+        (uint192 sellTokenPriceUSD, ) = sellAsset.price();
+        require(sellTokenPriceUSD != 0, "cannot price token");
```
tbrent, line 153: I understand the idea here, but I think one of the cases we want to account for (I didn't share this up front, my b) is when _we don't have a dust balance_, but _we think we do_. E.g, an oracle goes offline, `asset.price()` returns (0, FIX_MAX), but in fact we still hold millions of $ in the asset. 

I think supporting this usecase would mean forbidding the dutch auction TradeKind and only supporting batch auctions.

contracts/p1/BackingManager.sol
```diff
@@ -99,6 +99,84 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+
+        TradeRequest memory tradeRequest;
+        tradeRequest.sell = sellAsset;
+        tradeRequest.buy = assetRegistry.toAsset(IERC20(address(rToken)));
+        tradeRequest.sellAmount = sellQty;
+
+        // buyAmount = (sellQty * sellTokenPriceUSD) / rTokenPriceUSD
+        uint192 sellQtyInUSD = sellQty.mul(sellTokenPriceUSD, RoundingMode.CEIL);
```
tbrent, line 161: FYI you can just use `CEIL` here, it's an abbreviated version of `RoundingMode.CEIL`

contracts/p1/BackingManager.sol
```diff
@@ -99,6 +99,84 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
 
+    /// This will allow the callee to start a dust auction for a specific asset.
+    /// No other active trades must be open and current basket must be fully collateralized.
+    /// @param sellToken The asset we have a dust balance of
+    /// @custom:interaction RCEI
+    function startDustAuction(IERC20 sellToken, TradeKind kind) external notTradingPausedOrFrozen {
+        // @to reviewer: should we let the user set the TradeKind?
+
+        // == Refresh ==
+        // @to reviewer: Not sure what needs to be called/refreshed here
```
tbrent, line 110: If we buy the RToken, then we need both these lines. If we don't buy the RToken then we could get away with just refreshing the buy and sell asset.

contracts/p1/BackingManager.sol
tbrent: Maybe we can allow both TradeKinds, but revert if TradeKind is DUTCH_AUCTION and the price is 0

PR #809, opened 2023-05-05
3.0.0

tbrent:


tbrent:
frozen at tag [3.0.0-rc1](https://github.com/reserve-protocol/protocol/releases/tag/3.0.0-rc1)

tbrent:
unfrozen

tbrent:
@julianmrodri can you confirm the `.openzeppelin/mainnet-3_0_0.json` file is what we want?

julianmrodri:
> @julianmrodri can you confirm the `.openzeppelin/mainnet-3_0_0.json` file is correct and what we want?

The contents in that file match what it was deployed. However if you compile the code now, and pretend to reproduce the deployment the implementations of `BackingManagerP1` and `BasketHandlerP1`, produce a different hash. I double checked this and see there are no differences in terms of variables, slots, types, but some internal "references" to contracts differ a  bit. 

This does not create issues to upgrade or release new versions. I ran a POC locally for both contracts and was able to upgrade the 3.0.0 implementation with no issues. There was probably something different in the compilation at the time it was deployed vs now. But this is not a big deal.

I'll be opening a PR using Infura instead of Alchemy so we can group into a single network file.

tbrent:
> > @julianmrodri can you confirm the `.openzeppelin/mainnet-3_0_0.json` file is correct and what we want?
> 
> The contents in that file match what it was deployed. However if you compile the code now, and pretend to reproduce the deployment the implementations of `BackingManagerP1` and `BasketHandlerP1`, produce a different hash. I double checked this and see there are no differences in terms of variables, slots, types, but some internal "references" to contracts differ a bit (seems to be an internal id that OZ uses to connect compiled contracts)
> 
> This does not create issues to upgrade or release new versions. I ran a POC locally for both contracts and was able to upgrade the 3.0.0 implementation with no issues. There was probably something different in the compilation at the time it was deployed vs now. But this is not a big deal I think. I've opened a question with OZ to understand what could cause these differences.
> 
> I'll be opening a PR using Infura instead of Alchemy so we can group into a single network file.

Very nice, thanks.

tbrent:
It's also expected for `DeployerP1` to be different; I added an event to `deployRTokenAsset()` that isn't worth re-redeploying for.

julianmrodri:
The difference I was seeing was due to using different external library addresses which are linked to `BM` and `BH`. So this is perfectly fine if I use the same addresses you used for this deployment.

PR #811, opened 2023-05-08
upgrades phase 1 - prepare prior versions

julianmrodri:
* Prepare upgrades for future validations
* Creates `force-import` task
* Adds mainnet network file for versions `2.0.0`/`2.1.0`


tasks/upgrades/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    if (!deployments.implementations.components[keystr as keyof IComponents]) {
+      throw new Error(
+        `Missing deployed ${keystr} implementation for version ${version} in network ${hre.network.name}`
+      )
+    } else if (
+      !(await isValidContract(
+        hre,
+        deployments.implementations.components[keystr as keyof IComponents]
+      ))
+    ) {
```
pmckelvy1, line 52: else

tasks/upgrades/utils.ts
```diff
@@ -0,0 +1,58 @@
...
+    throw new Error(`TradingLib contract not found in network ${hre.network.name}`)
+  }
+
+  // Check Main implementation is defined
+  if (!deployments.implementations.main) {
+    throw new Error(
+      `Missing deployed Main implementation for version ${version} in network ${hre.network.name}`
+    )
+  } else if (!(await isValidContract(hre, deployments.implementations.main))) {
+    throw new Error(`TradingLib contract not found in network ${hre.network.name}`)
```
pmckelvy1, line 26: i think this error is wrong

pmckelvy1 (changes requested):

tasks/upgrades/utils.ts
julianmrodri: Yep should be Main will fix

tasks/upgrades/utils.ts
julianmrodri: Else is the expected behavior, just move forward

pmckelvy1:
lgtm

PR #812, opened 2023-05-08
Constant prime weights

tbrent:
a lot of tests got of whack because they assumed they could set a prime basket different from the one set in fixtures


test/fixtures.ts
```diff
@@ -408,7 +409,18 @@ export interface DefaultFixture extends RSRAndCompAaveAndCollateralAndModuleFixt
 
 type Fixture<T> = () => Promise<T>
 
+// Use this fixture when the prime basket will be constant at 1 USD
 export const defaultFixture: Fixture<DefaultFixture> = async function (): Promise<DefaultFixture> {
+  return await loadFixture(makeFixture.bind(null, true))
```
pmckelvy1, line 414: i think calling `loadFixture` with an anonymous function basically breaks the fixture. you need to set the result of `makeFixutre` to a const in order for it to "fixturize" it.  I don't think it functionally breaks anything, but the tests don't get the speed benefit of the fixturization

pmckelvy1: take a look at test/plugins/individual-collateral/fixture.ts .  We call `getDefaultFixture` and set it to variable in the `before` hook, before calling `loadFixture` in the `beforeEach` (ATokenFiatCollateral.test.ts & CTokenFiatCollateral.test.ts)

pmckelvy1 (changes requested):

test/fixtures.ts
pmckelvy1: after looking at other pipelines i guess this isnt the case...either im misreading this or this pattern doesn't affect speed? maybe because of the nested loadFixture? although i was reading some github issues last week that talked about how `loadFixture` does not work when nested, and they will not add support for it, so either that was an old thread or something about how we're doing this doesn't actually break `loadFixture`.  either way im a bit confused but if it works, it works...

test/Main.test.ts
```diff
@@ -1689,40 +1729,75 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
...
       await expect(
-        basketHandler.connect(owner).setPrimeBasket([token0.address], [0])
+        freshBasketHandler.connect(owner).setPrimeBasket([token0.address], [0])
       ).to.be.revertedWith('invalid target amount; must be nonzero')
+      await expect(
+        basketHandler.connect(owner).setPrimeBasket([token0.address], [0])
+      ).to.be.revertedWith('new basket missing target weights')
     })
 
     it('Should not allow to set prime Basket with RSR/RToken', async () => {
```
julianmrodri, line 1788: Maybe add the case of having all current Target names and amounts equal but adding a new Target name to see how its handled?

contracts/p1/BasketHandler.sol
```diff
@@ -534,8 +541,45 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
...
+                targetName,
+                contains ? amt + config.targetAmts[erc20] : config.targetAmts[erc20]
+            );
+        }
+
+        // Require new basket is exactly equal to old basket, in terms of targetAmts by targetName
+        len = newERC20s.length;
+        for (uint256 i = 0; i < len; ++i) {
+            bytes32 targetName = assetRegistry.toColl(newERC20s[i]).targetName();
+            // if the asset registry has a new registered asset targetName, this always reverts
```
julianmrodri, line 571: Its not clear if the comment is for the line on top or botton. Maybe remove the empty line? Also suggesting added a test case for this.

test/fixtures.ts
julianmrodri: So we are calling loadFixture twice? here and also in the tests? That seems confusing, not sure what is doing under the hood.

test/Recollateralization.test.ts
```diff
@@ -809,12 +816,6 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
         await stRSR.connect(addr1).stake(stakeAmount)
       })
 
-      afterEach(async () => {
```
julianmrodri, line 812: Why are we removing this? Seems like a check you dont want to drop.

test/fixtures.ts
tbrent: I didn't notice a difference in test execution speed or behavior under this approach, but I agree we should not use loadFixture if we can avoid it. Adding

contracts/p1/BasketHandler.sol
tbrent: I think I prefer having the comment below in this case. Having a clarifying comment first (as opposed to a descriptive or summary comment) makes the code less readable imo. I agree it's not totally clear what the comment applies to though, so I'll add some more detail to the text.

tbrent: I see though...this comment is super ambiguous in what it applies to. Removing and adding a test case.

test/Main.test.ts
tbrent: Yeah I haven't done new tests yet, adding today

test/Recollateralization.test.ts
tbrent: It's not something in 3.0.0 or master, fyi, it's something I added in `multiple-nonce-redemption`. I'm making some tweaks to keep it though, I see an easy way to do it now that doesn't involve changing tests to get them to end in a different state.

contracts/p1/BasketHandler.sol
```diff
@@ -644,5 +688,5 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[40] private __gap;
+    uint256[38] private __gap;
```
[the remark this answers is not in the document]
tbrent: Screenshot 2023-05-09 at 10 48 58 AM

Hmm there is something I don't understand...here's the output from `hardhat-storage-layout` plugin. The rightmost column is bytes; the 3rd column from the left is storage slot. It looks like `config` takes 4 slots; `basket` takes 2 slots; `_newBasket` takes 2 slots, and `_targetAmts` takes 3 slots. Is this wrong? (I added the c variable after _targetAmts, fyi). My prior understanding before looking at the results of this tool was also that structs / mappings take only 1 slot.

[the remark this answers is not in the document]
julianmrodri: Yes this might be right. Sorry, I looked incorrectly at the neext slot available.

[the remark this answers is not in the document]
julianmrodri: Then maybe should be 37? The plugins validation will capture this anyway.

tbrent:
Ok _now_ I believe this is complete.

test/Recollateralization.test.ts
julianmrodri: Yeah noticed later that it wasnt something we had before. Cool.

PR #814, opened 2023-05-09
Upgrades phase 2 - Deployment

julianmrodri:
* Upgrades phase 2 - Required tasks and scripts to check upgradeability for version 3.0.0
* Adds the `validate-upgrade` task to check at anytime via Mainnet fork if the current code is upgrade safe compared to a prior version.
* Modifies the implementations deployment script to use `upgrades.deployImplementation` or `upgrades.prepareUpgrade` to ensure we perform deployments via the OZ Upgrades plugins. Uses  a `LAST_DEPLOYED_VERSION` variable that needs to be set if this is an upgrade deployment.
* Modifies the format for the addresses JSON in `2.1.0` to be compatible with the new implementations (trade plugins change)

Note: Bumps hardhat and ethers version to take advantage of the latest functions in the OZ plugins.


julianmrodri:
One doubt I have is that maybe its better to use an `env` variable for `LAST_DEPLOYED_VERSION`?

Currently you need to set it at the top of the deployment script,  (for example to "2.1.0") if you are doing an upgrade. I dont like this a lot, but its not a task so I cannot set param. 

Looking forward to suggestions on this.

pmckelvy1:
lgtm

> One doubt I have is that maybe its better to use an `env` variable for `LAST_DEPLOYED_VERSION`?
> 
> Currently you need to set it at the top of the deployment script, (for example to "2.1.0") if you are doing an upgrade. I dont like this a lot, but its not a task so I cannot use a param.
> 
> What do you think?

i think this might be a good place to prompt the user.  that way we can 1) dynamically get the version number so that they don't need to remember to update the script, and 2) don't just run a fresh deployment if they forget to update the script w/ the version

i think something like this should work

```
export const prompt = async (query) => {
  const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout,
  });

  return new Promise((resolve) =>
    rl.question(query, (ans) => {
      rl.close();
      resolve(ans);
      return ans;
    })
  );
};

const ver = await prompt("enter the version number")
```

julianmrodri:
I can try that. how would that behave in CI though? Because we are running them in CI

julianmrodri:
> lgtm
> 
> > One doubt I have is that maybe its better to use an `env` variable for `LAST_DEPLOYED_VERSION`?
> > Currently you need to set it at the top of the deployment script, (for example to "2.1.0") if you are doing an upgrade. I dont like this a lot, but its not a task so I cannot use a param.
> > What do you think?
> 
> i think this might be a good place to prompt the user. that way we can 1) dynamically get the version number so that they don't need to remember to update the script, and 2) don't just run a fresh deployment if they forget to update the script w/ the version
> 
> i think something like this should work
> 
> ```
> export const prompt = async (query) => {
>   const rl = readline.createInterface({
>     input: process.stdin,
>     output: process.stdout,
>   });
> 
>   return new Promise((resolve) =>
>     rl.question(query, (ans) => {
>       rl.close();
>       resolve(ans);
>       return ans;
>     })
>   );
> };
> 
> const ver = await prompt("enter the version number")
> ```

I can try that. how would that behave in CI though? Because we are running them in CI

pmckelvy1:
export const prompt = async (query) => {
if (!useEnv('SKIP_PROMPT') {
  const rl = readline.createInterface({
    input: process.stdin,
    output: process.stdout,
  });

  return new Promise((resolve) =>
    rl.question(query, (ans) => {
      rl.close();
      resolve(ans);
      return ans;
    })
  );
} else {
return ''
}
};

const ver = await prompt("enter the version number")

pmckelvy1 (approved):
lgtm

PR #817, opened 2023-05-11
store erc20s in deployment file

julianmrodri:
* Stores `erc20s` addresses in the deployment files for all assets and collaterals.
* The address stored belongs to the main ERC20 used for that specific asset/collateral. This means it would be the wrapper tokens or vaults when they are applied.


tbrent:
LGTM!

PR #821, opened 2023-05-12
remove dependency on sol 0.4.24

pmckelvy1:


akshatmittal:
LGTM.

contracts/plugins/mocks/WETH.sol
```diff
@@ -30,7 +30,7 @@ contract WETH9 {
     mapping(address => uint256) public balanceOf;
     mapping(address => mapping(address => uint256)) public allowance;
 
-    function() external payable {
+    receive() external payable {
```
akshatmittal, line 33: This probably needs the fallback function.

PR #824, opened 2023-05-15
remove prime nonce

tbrent:
The primeNonce should have been deleted when we added `requireConstantConfigTargets()`

I checked the storage slots, and they are unchanged


pmckelvy1:
lgtm

PR #826, opened 2023-05-16
Correct vendor files

pmckelvy1:


tbrent:
Imports are broken elsewhere, but overall the change LGTM and we can merge after fixing

akshatmittal:
Possible to maybe not merge this until we move to the monorepo? Or be the first thing to be merged before any other changes so I can move this over.

tbrent:
> Possible to maybe not merge this until we move to the monorepo? Or be the first thing to be merged before any other changes so I can move this over.

The C4 coordinators already managed to count SLOC exempting these folders, but I think it's going to end up being important for the individual wardens during their audit that this code is cleanly separated out.

akshatmittal:
Yeah, we've already merged other things so might as well merge this. I'll move things together.

PR #827, opened 2023-05-17
Broker validations

julianmrodri:
* Adjusts broker validations and messages
* Updates tests
* New Min auction length = 2 blocks


tbrent:
LGTM

PR #829, opened 2023-05-18
Gas optimize rebalance

tbrent:
I already mentioned some context about this PR. In addition to that, I wanted to note that:
- The only part of the original caching I've kept around is the quantities[] array. I moved it into the primary context struct
- There's a new test for checking the RToken only gets sold after everything else
- There's new logic for preventing us from looking up the RToken price when the quantity is so low we can already tell it's not enough. I'm pretty sure the argument follows through, but it deserves a close eye.

### DIFF
--- a/contracts/p0/Distributor.sol
+++ b/contracts/p0/Distributor.sol
@@ -21,7 +21,7 @@ contract DistributorP0 is ComponentP0, IDistributor {
     address public constant FURNACE = address(1);
     address public constant ST_RSR = address(2);
 
-    uint8 public constant MAX_DESTINATIONS_ALLOWED = MAX_DESTINATIONS;
+    uint8 public constant MAX_DESTINATIONS_ALLOWED = 100;
 
     function init(IMain main_, RevenueShare memory dist) public initializer {
         __Component_init(main_);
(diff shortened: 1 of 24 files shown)


contracts/p0/Distributor.sol
```diff
@@ -21,7 +21,7 @@ contract DistributorP0 is ComponentP0, IDistributor {
     address public constant FURNACE = address(1);
     address public constant ST_RSR = address(2);
 
-    uint8 public constant MAX_DESTINATIONS_ALLOWED = 100;
+    uint8 public constant MAX_DESTINATIONS_ALLOWED = MAX_DESTINATIONS;
```
julianmrodri, line 24: Do we need MAX_DESTINATIONS_ALLOWED? Seems like both are the same concept and value?

julianmrodri (commented):
Looks good. Left a small comment.

test/RToken.test.ts
```diff
@@ -172,6 +172,27 @@ describe(`RTokenP${IMPLEMENTATION} contract`, () => {
...
+        await expect(rToken.connect(bmSigner).dissolve(fp('1'))).to.be.revertedWith(
+          'frozen or trading paused'
+        )
+
+        // Should revert if frozen
+        await main.connect(owner).unpauseTrading()
+        await main.connect(owner).freezeShort()
+        await expect(rToken.connect(bmSigner).dissolve(fp('1'))).to.be.revertedWith(
+          'frozen or trading paused'
+        )
```
pmckelvy1, line 192: unfreeze and call dissolve() just to directly confirm the positive case

pmckelvy1 (changes requested):

PR #831, opened 2023-05-20
3.0.0 audit tbrent

tbrent:
This PR represents my internal audit. I want to:
- Remove redundant notTradingPausedOrFrozen modifiers in `@custom:protected` functions
- Move modifiers in BackingManager to the top-level of the function where possible. In BackingManager's case this allows us to undo some RCEI structure that was costing gas. We had stuffed our modifiers into the Trading parent contract. In general it's dangerous to have nonReentrant anywhere other than first in the execution trace. Note that neither RCEI nor nonReentrant guards us from cross-contract reentrancy attacks; we have already accepted as a constraint of governance that ERC20 transfers should not pass execution. 
- Remove redundant notTradingPausedOrFrozen modifiers. Mainly: (i) Broker doesn't need it on openTrade and (ii) Distributor doesn't need it on distribute()
- Simplify BackingManager.forwardRevenue() when it is minting RTokens. It was doing two separate calls to `mint()` and `setBasketsNeeded()`, which was inefficient but more importantly obscured a very subtle bug where we were using the backingBuffer-bumped `needed` variable to calculate how many RTokens to mint, when this should have been left to RToken contract. The way it is now the RToken.mint() function handles basketsNeeded updates. 
- Make uniform _all_ RToken supply change logic. All supply changes go through the same 2 private helpers now: `_scaleUp()` and `_scaleDown()`. I'm down for different names, but I think this overall approach of pushing everything through the same BU-exchange-rate-preserving functions is good and we should stick with it. 
- I did an analysis of which functions in RToken could change the BU exchange rate and how. `melt()` and `setBasketsNeeded()` are the only ones at-risk of moving outside our [1e-9, 1e9] bounds. We had actually already done the analysis to figure out that `issue()/redeem()` were safe. Since we trust the furnace already not to `melt()` too fast, it only seems necessary to have the modifier on `setBasketsNeeded()`. I left it on _all_ the functions in the P0 implementation. Hopefully this doesn't cause problems for fuzzing.
- Change StRSR.cancelUnstake() modifier from notTradingPausedOrFrozen to notFrozen. StRSR.stake() is allowed while frozen or paused. For similar reasons we should allow cancelling an unstaking when paused. I'm hesitant to allow it when frozen as well because it's not a key part of the governance defense system, but we definitely can allow it when paused. 

Gas costs are pretty much better across the board. I think removing modifiers from RToken.melt() is worth it because of how many times it is going to get executed. It's part of all the important flows (issuance, redemption, trading). And moving exchangeRateIsValidAfter off issuance and redemption is going to provide massive savings in the long-run.

### DIFF
--- a/contracts/p1/RToken.sol
+++ b/contracts/p1/RToken.sol
@@ -471,6 +471,7 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
     /// @param recipient The address to receive the RTokens
     /// @param amtBaskets {BU} The number of amtBaskets to mint RToken for
     /// @param totalSupply {qRTok} The current totalSupply
+    // BU exchange rate cannot decrease, and it can only increase when < FIX_ONE.
     function _scaleUp(
         address recipient,
         uint192 amtBaskets,
@@ -491,6 +492,7 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
     /// @param account The address to dissolve RTokens from
     /// @param amount {qRTok} The amount of RToken to be dissolved
     /// @return baskets {BU} The equivalent number of baskets dissolved
+    // BU exchange rate cannot decrease, and it can only increase when < FIX_ONE.
     function _scaleDown(address account, uint256 amount) private returns (uint192 baskets) {
         // D18{BU} = D18{BU} * {qRTok} / {qRTok}
         baskets = basketsNeeded.muluDivu(amount, totalSupply()); // FLOOR
--- a/contracts/p1/RToken.sol
+++ b/contracts/p1/RToken.sol
@@ -181,7 +181,7 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
 
         assetRegistry.refresh();
         // solhint-disable-next-line no-empty-blocks
-        try main.furnace().melt() {} catch {}
+        try main.furnace().melt() {} catch {} // nice for the redeemer, but not necessary
 
         // == Checks and Effects ==
 
--- a/contracts/p1/RToken.sol
+++ b/contracts/p1/RToken.sol
@@ -113,14 +113,8 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
     /// @param recipient The address to receive the issued RTokens
     /// @param amount {qRTok} The quantity of RToken to issue
     /// @custom:interaction
-    // untestable:
-    //      `else` branch of `exchangeRateIsValidAfter` (ie. revert)
-    //       BU exchange rate cannot decrease, and it can only increase when < FIX_ONE.
-    function issueTo(address recipient, uint256 amount)
-        public
-        notIssuancePausedOrFrozen
-        exchangeRateIsValidAfter
-    {
+    // BU exchange rate cannot decrease, and it can only increase when < FIX_ONE.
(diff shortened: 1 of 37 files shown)


contracts/p1/RToken.sol
```diff
@@ -378,69 +337,72 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
-        external
-        notTradingPausedOrFrozen
-        exchangeRateIsValidAfter
-    {
+    function setBasketsNeeded(uint192 basketsNeeded_) external notTradingPausedOrFrozen {
         require(_msgSender() == address(backingManager), "not backing manager");
         emit BasketsNeededChanged(basketsNeeded, basketsNeeded_);
         basketsNeeded = basketsNeeded_;
+
+        // == P0 exchangeRateIsValidAfter modifier ==
```
pmckelvy1, line 394: P0

contracts/p1/RToken.sol
```diff
@@ -378,69 +337,72 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
+        uint256 supply = totalSupply();
+        require(supply > 0, "0 supply");
```
pmckelvy1, line 396: this seems weird at first glance, but i think its ok?  there should not be a situation where bm is calling setBasketsNeeded when there is 0 supply, as it only does so during a haircut.

contracts/p1/StRSR.sol
```diff
@@ -333,8 +334,7 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
     }
 
     function cancelUnstake(uint256 endId) external {
-        requireNotTradingPausedOrFrozen();
```
pmckelvy1, line 336: leaving a note to think about this some more...

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -45,15 +46,15 @@ contract DutchTrade is ITrade {
...
+    ITrading public origin; // the address that initialized the contract
 
     // === Auction ===
     IERC20Metadata public sell;
     IERC20Metadata public buy;
     uint192 public sellAmount; // {sellTok}
 
     // The auction runs from [startTime, endTime], inclusive
-    uint48 public startTime; // {s} when the dutch auction begins (1 block after init())
+    uint48 public startTime; // {s} when the dutch auction begins (12s after init())
```
pmckelvy1, line 57: if we don't yet have a single list of all the places that we consider 1 block == 12s, we should make one

pmckelvy1: it should probably go somewhere in the repo documentation

pmckelvy1 (changes requested):

contracts/p1/RToken.sol
tbrent: The comment is trying to say the below code is the exchangeRateIsValidAfter modifier from P0

contracts/p1/RToken.sol
tbrent: Previously we were returning in this case and not doing the following checks. That seemed wrong: to change `basketsNeeded` without doing any checks

contracts/plugins/trading/DutchTrade.sol
tbrent: Seems right. I'll create a file in `docs`

julianmrodri (commented):
Lgtm

PR #833, opened 2023-05-22
FacadeRead/Act: Parse version 

tbrent:

PR #834, opened 2023-05-22
3.0.0 Review Akshat

akshatmittal:
- Modifier order
- Redundant modifiers
- Versioning
- Comments
- Consistent naming

All minor nitpicks.

### DIFF
--- a/contracts/p1/BasketHandler.sol
+++ b/contracts/p1/BasketHandler.sol
@@ -20,6 +20,7 @@ import "./mixins/Component.sol";
 contract BasketHandlerP1 is ComponentP1, IBasketHandler {
     using BasketLibP1 for Basket;
     using EnumerableMap for EnumerableMap.Bytes32ToUintMap;
+    using EnumerableSet for EnumerableSet.Bytes32Set;
     using CollateralStatusComparator for CollateralStatus;
     using FixLib for uint192;
 
--- a/contracts/p1/StRSR.sol
+++ b/contracts/p1/StRSR.sol
@@ -158,8 +158,6 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
     uint48 private lastWithdrawRefresh; // {s} timestamp of last refresh() during withdraw()
     uint192 public withdrawalLeak; // {1} gov param -- % RSR that can be withdrawn without refresh
 
-    // NOTE for future: Do not add a uint48 (or smaller) here
-
     // ======================
 
     // init() can only be called once (initializer)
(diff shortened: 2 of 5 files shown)


contracts/p1/StRSR.sol
```diff
@@ -158,6 +158,8 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
     uint48 private lastWithdrawRefresh; // {s} timestamp of last refresh() during withdraw()
     uint192 public withdrawalLeak; // {1} gov param -- % RSR that can be withdrawn without refresh
 
+    // NOTE for future: Do not add a uint48 (or smaller) here
```
tbrent, line 161: Why? Seems fine to add. `withdrawalLeak` only takes up the first 20 bytes of the 32-byte slot, right?

contracts/p1/StRSR.sol
```diff
@@ -704,6 +706,10 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
     // solhint-disable-next-line no-empty-blocks
     function requireNotFrozen() private notFrozen {}
 
+    // contract-size-saver
+    // solhint-disable-next-line no-empty-blocks
+    function requireGovernanceOnly() private governance {}
```
tbrent, line 711: good change

contracts/p1/BasketHandler.sol
```diff
@@ -20,6 +20,7 @@ import "./mixins/Component.sol";
 contract BasketHandlerP1 is ComponentP1, IBasketHandler {
     using BasketLibP1 for Basket;
     using EnumerableMap for EnumerableMap.Bytes32ToUintMap;
+    using EnumerableSet for EnumerableSet.Bytes32Set;
```
tbrent, line 23: This is a nit on a nit, but we don't need this in BasketHandlerP1, only BasketLib where it already is

contracts/p1/StRSR.sol
tbrent: Is there a danger round upgrades when using a slot partially, that I don't know about?

akshatmittal: Storage layout is right-to-left, so it's preferred to isolate them between upgrades. (Although, it doesn't technically break it, just packs variables from two different versions)

contracts/p1/BasketHandler.sol
akshatmittal: Yeah, only because it makes it clear for someone reading it. Doesn't change the bytecode anyway.

contracts/p1/StRSR.sol
tbrent: I'm failing to see why right-to-left or left-to-right would have any bearing, as long as its consistent. 

On the other hand obviously I can see intuitively why there might be a preference for isolating slots between upgrades...but it seems like when you actually go down to the low-level it shouldn't matter

akshatmittal: Yeah, it most certainly a preference thing and not really a requirement per-se. I can drop this if you think this is unnecessary.

tbrent: I think I'd like to keep the `using Bytes32Set` change (clarifying) but remove this one comment (confusing)

tbrent: then LGTM for merge

tbrent: actually the tests will fail without an update to the testing files for 3.0.0, but that is going in as part of #833

PR #837, opened 2023-05-25
3.0.0 rc3

julianmrodri:
* Coverage for core contracts (some comments, minor tweaks, and tests)
* Coverage for plugins


scripts/verify_etherscan.ts
```diff
@@ -42,16 +42,20 @@ async function main() {
     '3_verify_deployer.ts',
     '4_verify_facade.ts',
     '5_verify_facadeWrite.ts',
-    '6_verify_collateral.ts',
```
pmckelvy1, line 45: undo

pmckelvy1 (changes requested):

PR #838, opened 2023-05-25
Updated plugin submission: Bizzyvinci bend dao

tbrent:


tbrent:
Still needs deployment scripts

PR #841, opened 2023-06-06
POC plugin tester

jankjr:


scripts/plugin_tester.ts
```diff
@@ -0,0 +1,103 @@
...
+    if (status === 2) {
+      break
+    }
+  }
+  return samples
+}
+
+export const main = async () => {
+  const PRICE_TIMEOUT = bn('604800') // 1 week
+  const REVENUE_HIDING = fp('0') // no revenue hiding by default; test individually
```
pmckelvy1, line 77: i think this makes sense, but it would be great to be able to run the simulation of a range of revenue hiding options and see how well the plugin does (0.01%, 0.001%, 0.0001%, etc...). maybe we take this functionality up 1 level and pass the revenueHiding variable in as an environment variable so we can run the same script in parallel in different threads

PR #842, opened 2023-06-07
dynamic network block time library

julianmrodri:
* Adds `NetworkConfiLib.sol` which returns the `blocktime()` for each network we support
* Adapts all contracts that were dependent on the 12s block time to use this library (P0 and P1). Basically adding constructors and immutable variables.
* Added Upgradeability and linting decorators to make process smooth for Tooling


pmckelvy1:
lgtm

pmckelvy1 (approved):

PR #843, opened 2023-06-07
run goerli deployment.

pmckelvy1:


tbrent:
What's the plan for this PR? I see there are some files in here that had to be touched just for the Goerli deployment, but aren't needed otherwise. We do value the new addresses, though.

tbrent (commented):
I forgot to hit submit before...

Look good but the rETH and stETH deployment scripts have mock oracle deployments which probably should come out

PR #848, opened 2023-06-15
Add tenderly plugin

pmckelvy1:


.github/workflows/tests.yml
```diff
@@ -18,6 +18,8 @@ jobs:
         with:
           node-version: 16.x
           cache: 'yarn'
+      - run: curl https://raw.githubusercontent.com/Tenderly/tenderly-cli/master/scripts/install-linux.sh | sudo sh
```
julianmrodri, line 21: Why do we need this on every task? I believe tests are still running on the Alchemy fork right?

julianmrodri (commented):
Getting this messages on `yarn compile`. I think it could use some additional documentation or comments on when Tenderly is used, and how to set it up and avoid this configuration warnings.

```
2023-06-19 17:11:50.348 WARN Service =>Tenderly config doesn't exist, empty string values are returned instead.
2023-06-19 17:11:50.382 WARN Service =>Tenderly config doesn't exist, empty string values are returned instead.
```

.github/workflows/tests.yml
pmckelvy1: it was a quick fix to silence some very loud tenderly logs that show up if the plugin is included in the hardhat config, but the user is not logged in via the cli. ill see if there is a better way to silence them

pmckelvy1: rolled back these additions and now the SILLY SERVICE logs seem to have disappeared...

PR #853, opened 2023-06-22
Update mattjurenka/morpho to 3.0.0

jankjr:


tbrent:
Hmm, so I think (i) the wrapper ERC20 is unnecessary; and (ii) it would actually break rewards. Rewards have to be done from off-chain, which is fine, but in this case they would just accumulate to the wrapper ERC20 instead of the BackingManager/RevenueTrader. We can get rid of the wrapper ERC20 by moving the `adjustExchangeRate()` function into the `refresh()` of the plugin contracts. 

The tasks are useful but we need deployment scripts too. The deployment scripts can use the task to do the deployment but the relevant thing is to commit to some set of constructor args, probably the same as the test suite but not necessarily. There's an example at `scripts/deployment/phase2-assets/collaterals/deploy_flux_finance_collateral.ts`.

contracts/plugins/assets/morpho-aave/MorphoAaveV2TokenisedDeposit.sol
```diff
@@ -0,0 +1,120 @@
...
+        (, , uint256 currentBalance) = morphoLens.getCurrentSupplyBalanceInOf(
+            poolToken,
+            address(this)
+        );
+        if (currentBalance == 0) {
+            return FIX_ONE;
+        }
+        return
+            shiftl_toFix(
+                _convertToShares(10**ERC20(underlying).decimals(), Math.Rounding.Down),
```
tbrent, line 116: minor optimization: I think it's worth making `ERC20(underlying).decimals()` an immutable var just given how many times this exchange rate is going to be computed

contracts/plugins/assets/morpho-aave/MorphoNonFiatCollateral.sol
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * @title MorphoNonFiatCollateral
+ * @notice Collateral plugin for a Morpho pool with non-fiat collateral, like WBTC
+ * Expected: {tok} != {ref}, {ref} is pegged to {target} unless defaulting, {target} != {UoA}
+ */
+contract MorphoNonFiatCollateral is MorphoFiatCollateral {
+    using OracleLib for AggregatorV3Interface;
+    using FixLib for uint192;
+
+    AggregatorV3Interface public immutable targetUnitChainlinkFeed; // {UoA/target}
```
tbrent, line 19: comment should be `{target/ref}`, I think, assuming this feed should be used for the `pegPrice`

package.json
```diff
@@ -23,6 +23,7 @@
     "test:p1": "PROTO_IMPL=1 hardhat test test/*.test.ts",
     "test:plugins": "hardhat test test/{libraries,plugins}/*.test.ts --parallel",
     "test:plugins:integration": "PROTO_IMPL=1 FORK=1 hardhat test test/plugins/individual-collateral/**/*.test.ts",
+    "test:maave": "PROTO_IMPL=1 FORK=1 hardhat test test/plugins/individual-collateral/morpho-aave/*.test.ts",
```
tbrent, line 26: don't need this, it'll get picked up by `test:plugins:integration`

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+  /******** Morpho collateral **************************/
+  const FiatCollateralFactory = await hre.ethers.getContractFactory(
+    "MorphoFiatCollateral"
+  )
+
+  {
+    const collateral = await FiatCollateralFactory.connect(deployer).deploy({
+      priceTimeout: priceTimeout.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDT!,
+      oracleError: fp('0.005').toString(), // 0.5%
```
tbrent, line 90: it's 0.25% over [here](https://docs.chain.link/data-feeds/price-feeds/addresses)

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+      erc20: maUSDT.address,
+      maxTradeVolume: fp('1e6').toString(), // $1m,
+      oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr
```
tbrent, line 93: and oracle timeout looks to be 86400s for USDT

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+
+    const collateral = await FiatCollateralFactory.connect(deployer).deploy({
+      priceTimeout: priceTimeout.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDC!,
+      oracleError: fp('0.005').toString(), // 0.5%
+      erc20: maUSDC.address,
+      maxTradeVolume: fp('1e6').toString(), // $1m,
+      oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr
+      targetName: ethers.utils.formatBytes32String("USD"),
+      defaultThreshold: fp('0.03').toString(), // 3%
```
tbrent, line 114: the rule we've been using for `defaultThreshold` is oracleError plus 1%, so 1.25% in this case

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+  }
+  {
+
+    const collateral = await FiatCollateralFactory.connect(deployer).deploy({
+      priceTimeout: priceTimeout.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDC!,
+      oracleError: fp('0.005').toString(), // 0.5%
+      erc20: maUSDC.address,
+      maxTradeVolume: fp('1e6').toString(), // $1m,
+      oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr
```
tbrent, line 112: should be 86400

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+    assetCollDeployments.collateral.maUSDT = collateral.address
+    deployedCollateral.push(collateral.address.toString())
+
+  }
+  {
+
+    const collateral = await FiatCollateralFactory.connect(deployer).deploy({
+      priceTimeout: priceTimeout.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDC!,
+      oracleError: fp('0.005').toString(), // 0.5%
```
tbrent, line 109: 0.25%

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+  {
+    const collateral = await FiatCollateralFactory.connect(deployer).deploy({
+      priceTimeout: priceTimeout.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDT!,
+      oracleError: fp('0.005').toString(), // 0.5%
+      erc20: maUSDT.address,
+      maxTradeVolume: fp('1e6').toString(), // $1m,
+      oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr
+      targetName: ethers.utils.formatBytes32String("USD"),
+      defaultThreshold: fp('0.03').toString(), // 3%
```
tbrent, line 95: 1.25%

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+      revenueHiding
+    );
+    assetCollDeployments.collateral.maUSDC = collateral.address
+    deployedCollateral.push(collateral.address.toString())
+  }
+  {
+    const collateral = await FiatCollateralFactory.connect(deployer).deploy({
+      priceTimeout: priceTimeout.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.DAI!,
+      oracleError: fp('0.005').toString(), // 0.5%
```
tbrent, line 126: 0.25%

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+      erc20: maDAI.address,
+      maxTradeVolume: fp('1e6').toString(), // $1m,
+      oracleTimeout: oracleTimeout(chainId, '3600').toString(), // 1 hr
```
tbrent, line 129: 3600 is right here, don't change it

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -0,0 +1,150 @@
...
+      targetName: ethers.utils.formatBytes32String("USD"),
+      defaultThreshold: fp('0.03').toString(), // 3%
```
tbrent, line 131: 1.25%

test/plugins/individual-collateral/morpho-aave/MorphoAAVEFiatCollateral.test.ts
```diff
@@ -0,0 +1,226 @@
...
+  collateralSpecificStatusTests,
+  beforeEachRewardsTest,
+  makeCollateralFixtureContext,
+  mintCollateralTo,
+  reduceTargetPerRef,
+  increaseTargetPerRef,
+  reduceRefPerTok,
+  increaseRefPerTok,
+  getExpectedPrice,
+  itClaimsRewards: it.skip,
```
tbrent, line 216: I think it's worth having a specific test for the proof-style claiming -- can probably grab a claim that occurred after this fork block and use that

test/plugins/individual-collateral/morpho-aave/constants.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const PRICE_TIMEOUT = bn(604800) // 1 week
+export const ORACLE_TIMEOUT = bn(86400) // 24 hours in seconds
+export const ORACLE_ERROR = fp('0.005')
+export const DEFAULT_THRESHOLD = bn(5).mul(bn(10).pow(16)) // 0.05
+export const DELAY_UNTIL_DEFAULT = bn(86400)
+
+export const USDC_DECIMALS = bn(6)
+
+export const FORK_BLOCK = 17528677
```
tbrent, line 17: there's another fork block number in `fork-block-numbers.ts`, not sure which is being used but there should probably only be 1

test/plugins/individual-collateral/morpho-aave/MorphoAAVENonFiatCollateral.test.ts
```diff
@@ -0,0 +1,261 @@
...
+} from './constants'
+import { MorphoAaveCollateralFixtureContext, mintCollateralTo } from './mintCollateralTo'
+
+interface MAFiatCollateralOpts extends CollateralOpts {
+  underlyingToken?: string
+  poolToken?: string
+  defaultPrice?: BigNumberish
+  defaultRefPerTok?: BigNumberish
+
+  refPerTokChainlinkFeed?: string
```
tbrent, line 35: all throughout this file I think the naming is refPerTok when it should be targetPerRef

contracts/plugins/assets/morpho-aave/MorphoNonFiatCollateral.sol
```diff
@@ -0,0 +1,62 @@
...
+    uint48 public immutable targetUnitOracleTimeout; // {s}
+
+    /// @param config Configuration of this collateral.
+    /// config.erc20 must be a IMorphoTokenisedDeposit
+    /// @param revenueHiding {1} A value like 1e-6 that represents the maximum refPerTok to hide
+    /// @param targetUnitChainlinkFeed_ Feed units: {UoA/target}
```
tbrent, line 25: units

contracts/plugins/assets/morpho-aave/MorphoAaveV2TokenisedDeposit.sol
```diff
@@ -0,0 +1,118 @@
...
+        underlyingOne = 10**ERC20(underlying).decimals();
+    }
+
+    // Offchain keepers should periodically call this such that we accumuate rewards
+    function claimOffchainRewards(uint256 _claimable, bytes32[] calldata _proof) external {
+        uint256 delta;
+        uint256 _totalSupply = totalSupply();
+        if (_totalSupply > 0) {
+            uint256 initialBal = rewardToken.balanceOf(address(this));
+            rewardsDistributor.claim(address(this), _claimable, _proof);
```
pmckelvy1, line 52: it looks like anyone can claim rewards for anyone else, which means someone could claim for this wrapper, locking the tokens in it.

if we use a storage variable instead of `initialBal`, we can reference the last reward balance in this calculation.  we would also want to utilize the `_claimAssetRewards()` function to update `rewardsPerShare`.

pmckelvy1 (changes requested):

contracts/plugins/assets/morpho-aave/MorphoAaveV2TokenisedDeposit.sol
tbrent: Isn't the idea that anyone can call this function, passing in the wrapper's address? Then in lines 59/60 they get their share of rewards? Maybe I'm misunderstanding something...

tbrent: The only args are the amount to claim and the proof

tbrent: (there could still be something wrong with the reward math I'm not seeing)

jankjr: @pmckelvy1 

@tbrent's assumption is correct. A keeper bot would have to generate a proof for the `MorphoAaveV2TokenisedDeposit` once in a while, and claim the rewards on behalf of the vault.

In this model, only the claimed Morpho tokens (via the `claimOffchainRewards`) method would be distributed to token holders.

jankjr: @pmckelvy1 I've added a test to show I intended Rewards to be claimed. Currently however rewards can actually claimed out of the Token vault, so they will just accumulate over time.

This is Morpho tokens are currently not transferable. Unless your address is whitelisted you're not allowed to transfer held tokens.

If we get Morpho token vault with a siginificant position we could ask the Morpho team nicely if they would be willing to whitelist the vault so users can claim their accumulated reward.

contracts/plugins/assets/morpho-aave/IMorpho.sol
```diff
@@ -0,0 +1,49 @@
...
+        external
+        view
+        returns (
+            uint256 balanceInP2P,
+            uint256 balanceOnPool,
+            uint256 totalBalance
+        );
+}
+
+abstract contract IMorphoTokenisedDeposit is RewardableERC4626Vault {
```
tbrent, line 31: Is there a reason to prefer this approach, as opposed to making the interface _just_ the `getExchangeRate()` function? It's also not an interface, despite being named as if it is. 

It would then be `is RewardableERC4626Vault, IMorphoTokenisedDeposit` when it comes to usage in the plugin, which is how we've done it elsewhere.

test/plugins/individual-collateral/morpho-aave/MorphoAAVENonFiatCollateral.test.ts
```diff
@@ -0,0 +1,260 @@
...
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const collateralSpecificConstructorTests = () => {}
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const collateralSpecificStatusTests = () => {}
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const beforeEachRewardsTest = async () => {}
+
+export const defaultCollateralOpts: MAFiatCollateralOpts = {
+  targetName: ethers.utils.formatBytes32String('WBTC'),
```
tbrent, line 217: This should be `BTC`. `WBTC` is the reference; `BTC` is the target

contracts/plugins/assets/morpho-aave/MorphoAaveV2TokenisedDeposit.sol
jankjr: @pmckelvy1 @tbrent I spoke with Patrick again. His concern is;

Since any can claim rewards for anyone in morpho. All they need to do is submit the proof and address. It's possible for someone to claim rewards that end up unaccounted.

I think this is a really good point, and is actually points to a more generic case that the`RewardableERC20` does not handle. What happens if reward tokens are just transfered to the token vault outside of us claiming them? These tokens would currently not be accounted for. On low fee chains, it's is not that unusual to see direct airdrops from protocols, and with Morpho, you don't have to be the one initiating a claim to receive rewards.

I'll make a small modification to the `RewardableERC20` that should handle it. This actually simplifies the Morpho implementation a lot.

contracts/plugins/assets/morpho-aave/MorphoTokenisedDeposit.sol
```diff
@@ -0,0 +1,85 @@
...
+        SafeERC20.safeTransferFrom(IERC20(underlying), caller, address(this), assets);
+        SafeERC20.safeApprove(IERC20(underlying), address(morphoController), assets);
+        morphoController.supply(poolToken, assets);
+
+        _mint(receiver, shares);
+        emit Deposit(caller, receiver, assets, shares);
+    }
+
+    function _decimalsOffset() internal view virtual override returns (uint8) {
+        return 0;
```
tbrent, line 67: I think this is right, but just want to check: normally when dealing with 4626 vaults we use a non-zero decimal offset in order to prevent [this inflation attack](https://mixbytes.io/blog/overview-of-the-inflation-attack).

In this case though we have a non-transferrable asset (underlying). Nobody can transfer the underlying, because it's a non-transferrable position to begin with. I think this means we can know the inflation attack cannot be performed, but I'm not sure. 

The other direction possible would be manipulating the totalSupply downwards, instead of the totalAssets upwards. I don't think this is possible either, but it's not super obvious so I wanted to mention all of this.

contracts/plugins/assets/morpho-aave/MorphoSelfReferentialCollateral.sol
```diff
@@ -0,0 +1,67 @@
...
+import { MorphoTokenisedDeposit } from "./MorphoTokenisedDeposit.sol";
+import { OracleLib } from "../OracleLib.sol";
+// solhint-disable-next-line max-line-length
+import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import { shiftl_toFix, FIX_ONE, FixLib, CEIL } from "../../../libraries/Fixed.sol";
+
+/**
+ * @title MorphoSelfReferentialCollateral
+ * @notice Collateral plugin for a Morpho pool with self referential collateral, like stETH
+ * Expected: {tok} != {ref}, {ref} is pegged to {target} unless defaulting, {target} == {UoA}
```
tbrent, line 15: should be `{target} != {UoA}`

test/plugins/RewardableERC20.test.ts
```diff
@@ -154,6 +155,17 @@ for (const wrapperName of wrapperNames) {
           expect(await rewardableVault.rewardsPerShare()).to.equal(bn(0))
           expect(await rewardableVault.lastRewardsPerShare(alice.address)).to.equal(bn(0))
         })
+
+        it('supports direct airdrops', async () => {
```
tbrent, line 159: great addition

contracts/plugins/assets/morpho-aave/MorphoAaveV2TokenisedDeposit.sol
tbrent: Ah, @pmckelvy1 good catch. Yes. Need storage var.

tbrent:
@pmckelvy1 Jan's most recent update to `RewardableERC20` seems like it handles rewards right to me, but you're more familiar than I am with this and should double-check me

contracts/plugins/assets/erc20/RewardableERC20.sol
```diff
@@ -24,6 +26,11 @@ abstract contract RewardableERC20 is IRewardable, ERC20 {
     mapping(address => uint256) public accumulatedRewards; // {qRewards}
     mapping(address => uint256) public claimedRewards; // {qRewards}
 
+    // Used to keep track of how many reward the Vault has accumulated
+    // Whenever _claimAndSyncRewards() is called we will calculate the difference
+    // between the current balance and `previousBalance` to figure out how much to distribute
+    uint256 internal previousBalance = 0;
```
pmckelvy1, line 32: i think `lastRewardBalance` is a bit more explicit

contracts/plugins/assets/erc20/RewardableERC20.sol
```diff
@@ -36,39 +43,59 @@ abstract contract RewardableERC20 is IRewardable, ERC20 {
...
-        lastRewardsPerShare[account] = rewardsPerShare;
-        accumulatedRewards[account] += newRewards;
+        uint256 shares = balanceOf(account);
+        uint256 _accumuatedRewards = accumulatedRewards[account];
+        uint256 _rewardsPerShare = rewardsPerShare;
+        if (accountRewardsPerShare < _rewardsPerShare) {
+            uint256 delta = _rewardsPerShare - accountRewardsPerShare;
+            delta = (delta * shares) / one;
+            // {qRewards} = {qRewards/share} * {qShare} / {qShare/share}
+            _accumuatedRewards = _accumuatedRewards + delta;
```
pmckelvy1, line 60: to avoid changing the units of `delta`, lets just combine these 2 lines into
```
// {qRewards} = {qRewards} + {qRewards/share} * {qShare} / {qShare/share}
_accumulatedRewards += (delta * shares) / one;
```

contracts/plugins/assets/erc20/RewardableERC20.sol
```diff
@@ -36,39 +43,59 @@ abstract contract RewardableERC20 is IRewardable, ERC20 {
...
     }
 
     function _claimAccountRewards(address account) internal {
         uint256 claimableRewards = accumulatedRewards[account] - claimedRewards[account];
+        if (claimableRewards == 0) {
+            return;
+        }
+
         emit RewardsClaimed(IERC20(address(rewardToken)), claimableRewards);
-        if (claimableRewards == 0) return;
```
pmckelvy1, line 69: i guess this is up for debate but i think we always want to emit `RewardsClaimed` even if the amount is 0. @tbrent thoughts?

contracts/plugins/assets/morpho-aave/MorphoTokenisedDeposit.sol
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    function _withdraw(
```
pmckelvy1, line 70: this function doesnt seem to actually be sending any tokens to the withdrawer

contracts/plugins/assets/morpho-aave/MorphoTokenisedDeposit.sol
```diff
@@ -0,0 +1,85 @@
...
+
+struct MorphoTokenisedDepositConfig {
+    IMorpho morphoController;
+    IMorphoRewardsDistributor rewardsDistributor;
+    IERC20Metadata underlyingERC20;
+    IERC20Metadata poolToken;
+    ERC20 rewardToken;
+}
+
+abstract contract MorphoTokenisedDeposit is RewardableERC4626Vault {
```
pmckelvy1, line 17: we should write a unit test suite specifically for this contract

pmckelvy1 (changes requested):

contracts/plugins/assets/morpho-aave/MorphoTokenisedDeposit.sol
jankjr: Oh wow, that's some oversight 🤦

contracts/plugins/assets/morpho-aave/MorphoTokenisedDeposit.sol
jankjr: I will add a suite that tests withdrawals and deposits for all of Morphos markets.

contracts/plugins/assets/erc20/RewardableERC20.sol
jankjr: I moved up above the if statement.

contracts/plugins/assets/erc20/RewardableERC20.sol
jankjr: Good point

PR #855, opened 2023-06-29
Add script to sync signatures with 4bytes

jankjr:
To run the signature sync: `yarn run:4bytes`.

Script will pull all event/function signatures from artifacts/ and sync them with 4bytes if we've not previously syncced that signature with 4bytes.

The `scripts/4bytes-syncced.json` file has been added to the repo, this file contains all the signatures we've previously syncced. The file is updated as `yarn run:4bytes` works through the signatures.

If you run the script locally, be sure to commit the `scripts/4bytes-syncced.json` to history.

I also added a workflow that should trigger whenever we merge to master. This workflow will trigger the sync file and commit it to history when done.


pmckelvy1 (approved):
lgtm

PR #856, opened 2023-06-29
Add multiple pausers/freezers to 2.1.0 FacadeWrite

tbrent:


julianmrodri (approved):
We should wait till tests run before merging

julianmrodri:
There is a test for CTokenFiatCollateral that needs to be fixed

PR #863, opened 2023-07-05
Stargate Asset Plugin

akshatmittal:


contracts/plugins/assets/stargate/StargatePoolFiatCollateral.sol
```diff
@@ -0,0 +1,121 @@
...
+        CollateralStatus newStatus = status();
+        if (oldStatus != newStatus) {
+            emit CollateralStatusChanged(oldStatus, newStatus);
+        }
+    }
+
+    /// @return _rate {ref/tok} Quantity of whole reference units per whole collateral tokens
+    function refPerTok() public view virtual override returns (uint192 _rate) {
+        IStargatePoolWrapper poolWrapper = IStargatePoolWrapper(address(erc20));
+        IStargatePool pool = IStargatePool(address(poolWrapper.pool()));
```
tbrent, line 109: Since it's immutable we should probably store `poolWrapper.pool()` and avoid re-lookup

tbrent (commented):
I still have more to go through, but up-front: did you consider inheriting the plugin from `AppreciatingFiatCollateral`? I think it would allow you to skip re-implementing `tryPrice()` and `refresh()`, and `tryPrice()`. I guess you know no revenue hiding is required, so this way is a gas-saver? Still, I think I'd make the other tradeoff. Maybe there's something I'm not considering though...interested if you had more thoughts. 

Similarly, what about inheriting from `RewardableERC4626Vault` for the wrapper? It might not be a good fit, but wanted to check.

akshatmittal:
@tbrent Yeah I did consider the `AppreciatingFiatCollateral` route, but didn't do it because it didn't require revenue hiding and I intended to keep as much of the original submitted plugin as I can.

contracts/plugins/assets/stargate/StargatePoolFiatCollateral.sol
```diff
@@ -0,0 +1,121 @@
...
+        uint256 _totalSupply = pool.totalSupply();
```
pmckelvy1, line 110: should we do the shift here instead? in the odd case that erc20Decimals is > 18, the shift may put _totalSupply at 0

pmckelvy1:
+1 for using `RewardableERC4626Vault` or `RewardableERC20Wrapper`, it looks like one of those should work (probably 4626

pmckelvy1 (changes requested):

contracts/plugins/assets/stargate/StargatePoolFiatCollateral.sol
tbrent: We should not do the shift and instead use `Fixed.divuu()` directly

akshatmittal: Yeah I thought so. I'll update this.

PR #865, opened 2023-07-07
fix storage slots Distributor

julianmrodri:
We had an issue with gap in the Distributor, as 2 new variables were added: rsrTrader and rTokenTrader taking up 2 slots.


tbrent:
Confirming this was needed

PR #867, opened 2023-07-11
Base Testnet configuration

julianmrodri:
* Configuration and scripts for deployment and verification on Base Goerli (Testnet)
* Includes deployed addresses like we have for Eth Goerli


tbrent:
LGTM

PR #880, opened 2023-08-03
gas test asset.refresh() for all tests

tbrent:


test/plugins/Asset.test.ts
```diff
@@ -102,7 +114,7 @@ describe('Assets contracts #fast', () => {
       rTokenAsset,
     } = await loadFixture(defaultFixture))
 
-    // Get collateral tokens
+    // Get rsrAsset tokens
```
julianmrodri, line 117:
```suggestion
    // Get collateral tokens
```

julianmrodri (approved):
LGTM! There is a small comment where it says "rsrAsset" instead of "collateral". I think that was a mistake.

PR #881, opened 2023-08-07
Stargate fixes

pmckelvy1:


akshatmittal:
I'd also recommend looking at the deployment parameters. We likely need to tighten them here.

pmckelvy1:
> I'd also recommend looking at the deployment parameters. We likely need to tighten them here.

what do you think we should tighten?

akshatmittal:
> what do you think we should tighten?

Have a look at the peg parameters and error settings.

akshatmittal:
lgtm

PR #884, opened 2023-08-08
C4: add test to show effect of someone frontrunning a gnosis trade

jankjr:
Add a simple test to make sure GnosisTrades can be settled even if someone manupuated tokens bought.

Added a small comment that explains the effect and reasons for not fixing it.

### DIFF
--- a/test/Broker.test.ts
+++ b/test/Broker.test.ts
@@ -22,6 +22,7 @@ import {
   GnosisMock,
   GnosisMockReentrant,
   GnosisTrade,
+  GnosisTrade__factory,
   IAssetRegistry,
   TestIBackingManager,
   TestIBasketHandler,
@@ -55,6 +56,7 @@ import {
 } from './utils/time'
 import { ITradeRequest } from './utils/trades'
 import { useEnv } from '#/utils/env'
+import { parseUnits } from 'ethers/lib/utils'
 
 const DEFAULT_THRESHOLD = fp('0.01') // 1%
 const DELAY_UNTIL_DEFAULT = bn('86400') // 24h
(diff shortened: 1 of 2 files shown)


test/Broker.test.ts
```diff
@@ -854,6 +854,62 @@ describe(`BrokerP${IMPLEMENTATION} contract #fast`, () => {
         expect(await trade.canSettle()).to.equal(false)
       })
 
+      it('Settle frontrun regression check - should be OK', async () => {
+        // Initialize trade - simulate from backingManager
+        const tradeRequest: ITradeRequest = {
+          sell: collateral0.address,
+          buy: collateral1.address,
+          sellAmount: amount,
+          minBuyAmount: bn('0'),
```
pmckelvy1, line 863: probably best to initialize this with something realistic (> 0), since 0 won't ever happen

test/Broker.test.ts
```diff
@@ -854,6 +854,62 @@ describe(`BrokerP${IMPLEMENTATION} contract #fast`, () => {
...
+        await whileImpersonating(backingManager.address, async (bmSigner) => {
+          const tradeWithBm = trade.connect(bmSigner)
+          const normalValues = await tradeWithBm.callStatic.settle()
+
+          // Simulate someone frontrunning settlement and adding more traded funds to Trade
+          await token0.connect(owner).mint(trade.address, amount.sub(1))
+
+          const frontRunnedValues = await tradeWithBm.callStatic.settle()
+
+          expect(normalValues.soldAmt).to.not.eq(frontRunnedValues.soldAmt)
```
pmckelvy1, line 903: let's also validate that the boughtAmt was correct

pmckelvy1 (changes requested):

PR #885, opened 2023-08-08
C4 #29

tbrent:
Resolves https://github.com/code-423n4/2023-06-reserve-findings/issues/29

### DIFF
--- a/test/Furnace.test.ts
+++ b/test/Furnace.test.ts
@@ -452,9 +452,15 @@ describe(`FurnaceP${IMPLEMENTATION} contract`, () => {
 
       // Transfer to Furnace and do first melt
       await rToken.connect(addr1).transfer(furnace.address, bn('10e18'))
+      await setNextBlockTimestamp(Number(await getLatestBlockTimestamp()) + Number(ONE_PERIOD))
       await furnace.melt()
 
-      // Advance 99 periods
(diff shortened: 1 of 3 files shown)


test/Furnace.test.ts
```diff
@@ -446,6 +446,34 @@ describe(`FurnaceP${IMPLEMENTATION} contract`, () => {
...
+        .to.emit(furnace, 'RatioSet')
+        .withArgs(config.rewardRatio, maxRatio)
+
+      // Unfreeze and advance 1 period
+      await main.connect(owner).unfreeze()
+      await setNextBlockTimestamp(Number(await getLatestBlockTimestamp()) + Number(ONE_PERIOD))
+      await expect(furnace.melt()).to.emit(rToken, 'Melted')
+
+      // Should have updated lastPayout + lastPayoutBal
+      expect(await furnace.lastPayout()).to.be.closeTo(await getLatestBlockTimestamp(), 12)
```
pmckelvy1, line 473: don't we want to check these after the setRatio call and before the last melt as well?

pmckelvy1 (changes requested):

test/Furnace.test.ts
tbrent: Absolutely, yes. I was focused on the 99 blocks happening at the old rate, but I missed doing the check directly as well.

PR #886, opened 2023-08-09
C4: handle Chainlink oracle deprecation

jankjr:
If a chainlink oracle gets deprecated, the aggregator address will be set to 0.

This PR extends OracleLib to correctly handle this case, and reverting with StalePrice.

### DIFF
--- a/contracts/plugins/mocks/ChainlinkMock.sol
+++ b/contracts/plugins/mocks/ChainlinkMock.sol
@@ -23,6 +23,7 @@ contract MockV3Aggregator is AggregatorV3Interface {
 
     // Additional variable to be able to test invalid behavior
     uint256 public latestAnsweredRound;
+    address public aggregator;
 
     mapping(uint256 => int256) public getAnswer;
     mapping(uint256 => uint256) public getTimestamp;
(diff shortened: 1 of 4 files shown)


contracts/plugins/mocks/ChainlinkMock.sol
```diff
@@ -102,13 +114,7 @@ contract MockV3Aggregator is AggregatorV3Interface {
...
         )
     {
-        return (
-            uint80(latestRound),
-            getAnswer[latestRound],
-            getStartedAt[latestRound],
-            getTimestamp[latestRound],
-            uint80(latestAnsweredRound)
-        );
+        return this.getRoundData(uint80(latestAnsweredRound));
```
pmckelvy1, line 117: this is changing the response we normally get here (using `latestAnsweredRound` for `roundId` instead of `latestRound`).  all the tests are passing so it doesnt seem to have an effect, but we should probably still keep it as it was

pmckelvy1 (changes requested):

tbrent:
https://github.com/code-423n4/2023-06-reserve-findings/issues/8

PR #890, opened 2023-08-10
RecollateralizationLib gas optimization

tbrent:
Couple things happening in this PR
1. Fix Recollateralization gas tests to actually run successfully
2. Update gas snapshots to show where `3.0.0-rc5` actually was
3. Apply gas optimizations to `RecollateralizationLibP1.sol` to reduce the need for a mulDiv for every single asset, on the range.bottom side.
4. Compute new gas snapshots

So the total diff shown in the PR is the result of (i) a large increase that was already present in `3.0.0-rc5` and (ii) a moderate decrease that I applied by gas optimizing the range.bottom computation. Attaching a screenshot of _just_ that decrease so it can be easily seen, but you could also go [here](https://github.com/reserve-protocol/protocol/commit/392997d51002d45d249cddab72522356553d45be)

Screenshot 2023-08-10 at 6 56 11 PM

### DIFF
--- a/contracts/p1/mixins/RecollateralizationLib.sol
+++ b/contracts/p1/mixins/RecollateralizationLib.sol
@@ -181,6 +181,9 @@ library RecollateralizationLibP1 {
         int256 deltaTop; // D18{BU} even though this is int256, it is D18
         // not required for range.bottom
 
+        // to minimize total operations, range.bottom is calculated from a summed UoA
+        uint192 uoaBottom; // {UoA} pessimistic UoA estimate of balances above basketsHeld.bottom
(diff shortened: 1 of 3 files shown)


akshatmittal (approved):

pmckelvy1:
changes look sound, but looks like dutch trade costs went up a little bit?

pmckelvy1:
> changes look sound, but looks like dutch trade costs went up a little bit?

nvm i understand now that those don't reference the correct starting gas

pmckelvy1 (approved):
lgtm

PR #892, opened 2023-08-11
Optimize Events

akshatmittal:


contracts/interfaces/IFacadeWrite.sol
```diff
@@ -84,16 +84,13 @@ interface IFacadeWrite {
     /// @param rToken The address of the RToken
     /// @param governance The address of the new governance
     /// @param timelock The address of the timelock
-    event GovernanceCreated(
-        IRToken indexed rToken,
-        address indexed governance,
-        address indexed timelock
-    );
+    event GovernanceCreated(IRToken rToken, address governance, address timelock);
```
tbrent, line 87: I think we should keep all 3 indexed here, since this only runs once per RToken instance.

contracts/interfaces/IDistributor.sol
```diff
@@ -28,13 +28,13 @@ interface IDistributor is IComponent {
     /// @param dest The address set to receive the distribution
     /// @param rTokenDist The distribution of RToken that should go to `dest`
     /// @param rsrDist The distribution of RSR that should go to `dest`
-    event DistributionSet(address dest, uint16 rTokenDist, uint16 rsrDist);
+    event DistributionSet(address indexed dest, uint16 rTokenDist, uint16 rsrDist);
```
tbrent, line 31: Very nice. Was going to comment that I think we should add it here, but you already saw that.

contracts/interfaces/IMain.sol
```diff
@@ -114,43 +114,43 @@ interface IAuth is IAccessControlUpgradeable {
 interface IComponentRegistry {
     // === Component setters/getters ===
 
-    event RTokenSet(IRToken indexed oldVal, IRToken indexed newVal);
+    event RTokenSet(IRToken oldVal, IRToken newVal);
```
tbrent, line 117: Huh, looks like these events are leftover from a time where it was possible to change the addresses, which is no longer the case. I think it's probably not worth changing, though. You could imagine an RToken upgrades Main to include the ability to change the component addresses, so maybe it's worth keeping around. And the cost is basically zero.

contracts/interfaces/IStRSR.sol
```diff
@@ -83,19 +83,19 @@ interface IStRSR is IERC20MetadataUpgradeable, IERC20PermitUpgradeable, ICompone
...
-    event ExchangeRateSet(uint192 indexed oldVal, uint192 indexed newVal);
+    event ExchangeRateSet(uint192 oldVal, uint192 newVal);
 
     /// Emitted whenever RSR are paids out
-    event RewardsPaid(uint256 indexed rsrAmt);
+    event RewardsPaid(uint256 rsrAmt);
 
     /// Emitted if all the RSR in the staking pool is seized and all balances are reset to zero.
-    event AllBalancesReset(uint256 indexed newEra);
+    event AllBalancesReset(uint256 newEra);
```
tbrent, line 92: Let's leave indexed here since it runs so infrequently and it might be useful to query

contracts/interfaces/IStRSR.sol
```diff
@@ -83,19 +83,19 @@ interface IStRSR is IERC20MetadataUpgradeable, IERC20PermitUpgradeable, ICompone
...
     /// Emitted if all the RSR in the unstakin pool is seized, and all ongoing unstaking is voided.
-    event AllUnstakingReset(uint256 indexed newEra);
+    event AllUnstakingReset(uint256 newEra);
```
tbrent, line 94: same

tbrent (commented):
Nice, I agree with almost all these choices, but have a few places I think it's worth doing something different. 

Also something may have happened with your IDE lint config, take a look.

akshatmittal:
Closing in favor of #894

PR #894, opened 2023-08-11
Bunch of gas optimizations

akshatmittal:
1. Resolves https://github.com/code-423n4/2023-06-reserve-findings/issues/27
2. Resolves https://github.com/code-423n4/2023-06-reserve-findings/issues/50
3. Optimizes Events


pmckelvy1 (approved):
lgtm

pmckelvy1:
can you post a before/after gas diff?

contracts/p1/BasketHandler.sol
```diff
@@ -255,7 +255,12 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
 
         for (uint256 i = 0; i < size; ++i) {
             CollateralStatus s = assetRegistry.toColl(basket.erc20s[i]).status();
-            if (s.worseThan(status_)) status_ = s;
+            if (s.worseThan(status_)) {
+                status_ = s;
+            }
+            if (status_ == CollateralStatus.DISABLED) {
```
tbrent, line 261: Not sure you saw my comment in the other PR but I think this change will end up being a net gas increase given how infrequently one of the backing will be DISABLED.

A change I like that might be the best of both worlds is putting this check inside the if statement above

akshatmittal: You are right, I was going to remove this, just missed it.

tbrent:
As expected, the gas improvements are extremely marginal

PR #897, opened 2023-08-15
c4: #5 Potential Early Exploit in Morho-Aave ERC4626 Implementation

jankjr:
Findings:

It **is** possible to inflate the value of a single share by manupuating the amount of tokens the vault holds.

But I don't think it is possible to do this in a profitable manner - as I can't see anyway to execute the attack in a way where the attacker does not loose funds. The issue comes from our implementation of the ERC4626 essentially adding an additional share to the vault. This zero'th share will end up holding the profit of the attack with no way to redeem it.

The inflation attack was fixed by setting the decimalOffset to 9. I added a test that will fail if the inflation attack manifests again.

### DIFF
--- a/test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
+++ b/test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
@@ -252,7 +252,7 @@ const execTestForToken = ({
       await closeTo(methods.assets(bob), fraction(50))
     })
 
-    it('Check that inflation attacks are not possible', async () => {
+    it('Regression Test - C4 July 2023 Issue #5', async () => {
       const {
         users: { alice, bob },
         methods,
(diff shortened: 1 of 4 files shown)


tbrent:
if I'm supposed to review this you should tag me as a reviewer otherwise I won't get the notification

test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
```diff
@@ -212,11 +252,57 @@ const execTestForToken = ({ token, poolToken, amount }: ITestSuiteVariant) => {
       await closeTo(methods.assets(bob), fraction(50))
     })
 
+    it('Check that inflation attacks are not possible', async () => {
```
tbrent, line 255: Let's use the same pattern from here when it comes to naming / documentation: https://github.com/reserve-protocol/protocol/blob/11aa80b8235689484c84b452c7e1145fcef988d7/test/Furnace.test.ts#L450

tbrent (commented):
Looks good, just a small comment on test naming

pmckelvy1:
fixes https://github.com/code-423n4/2023-07-reserve-findings/issues/5

pmckelvy1:
https://github.com/code-423n4/2023-07-reserve-findings/issues/5

PR #903, opened 2023-08-21
Restrict disabling dutch auctions to those started by the BackingManager

tbrent:
Reason for the change is that RevenueTrader does not respect `minTradeVolume` (intentional), but this could allow someone to trigger an auction for ~$0 in tokens and bid in the geometric phase in order to prevent the BackingManager from rebalancing via dutch auction.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+# 3.1.0 - Unreleased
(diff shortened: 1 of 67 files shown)


tbrent:
@jankjr tagging you in on this one

akshatmittal:
lgtm!

PR #906, opened 2023-08-21
Optional price caching

tbrent:
Look like it would decrease gas costs for all but one rebalancing auction, which is the final bid in a sequence of rebalances. Revenue auctions slightly more expensive but not meaningfully


tbrent:
Currently leaning towards closing this and not including the change

tbrent:
Closing since not meaningful enough

PR #910, opened 2023-08-22
update Asset price comment.

pmckelvy1:


pmckelvy1:
documents https://github.com/code-423n4/2023-07-reserve-findings/issues/24

contracts/plugins/assets/Asset.sol
```diff
@@ -112,6 +112,12 @@ contract Asset is IAsset, VersionedAsset {
     /// @dev Should be general enough to not need to be overridden
     /// @return _low {UoA/tok} The lower end of the price estimate
     /// @return _high {UoA/tok} The upper end of the price estimate
+    /// @notice If the price feed is broken, _low will decay downwards and _high will decay upwards
+    ///     If the price feed is broken for more than `oracleTimeout + priceTimeout` seconds,
+    ///     _low will be 0 and _high will be FIX_MAX.
+    ///     Because the price decay begins at `oracleTimeout` seconds and not `updateTime` from the
+    ///     price feed, the price feed can be broken for up to `priceTimeout` seconds without
```
tbrent, line 119: I think it's supposed to be "this price feed can be broken for up to `~2 * oracleTimeout` seconds without affecting the price estimate". It's oracleTimeout and not priceTimeout, and it's ~2x because the price feed can break and someone can refresh() just before the first oracleTimeout, and then another whole oracleTimeout would need to pass after before the decay starts.

contracts/plugins/assets/Asset.sol
```diff
@@ -112,6 +112,14 @@ contract Asset is IAsset, VersionedAsset {
     /// @dev Should be general enough to not need to be overridden
     /// @return _low {UoA/tok} The lower end of the price estimate
     /// @return _high {UoA/tok} The upper end of the price estimate
+    /// @notice If the price feed is broken, _low will decay downwards and _high will decay upwards
+    ///     If the price feed is broken for more than `oracleTimeout + priceTimeout` seconds,
```
tbrent, line 116: Small nit I can make myself: Technically this isn't true due to the 2x part mentioned below, so really it should be "If `price()` is broken for more than..."

PR #914, opened 2023-08-24
save gas by only checking aggregator if there is a revert

tbrent:
https://github.com/code-423n4/2023-08-reserve-mitigation-findings/issues/28


akshatmittal:
Looks good, can we create a test case to verify it works? Possibly run it against a deprecated oracle?

tbrent:
We have one: https://github.com/reserve-protocol/protocol/blob/c4-28/test/plugins/OracleDeprecation.test.ts

akshatmittal:
Ah, forgot about that one. Looks good. LGTM.

akshatmittal (approved):

PR #917, opened 2023-08-25
C4 +inf

tbrent:
- https://github.com/code-423n4/2023-07-reserve-findings/issues/20
- https://github.com/code-423n4/2023-07-reserve-findings/issues/21

### DIFF
--- a/contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
+++ b/contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
@@ -92,7 +92,7 @@ contract CurveStableMetapoolCollateral is CurveStableCollateral {
 
         // {UoA/tok} = {UoA} / {tok}
         low = aumLow.div(supply, FLOOR);
-        high = aumHigh != FIX_MAX ? aumHigh.div(supply, CEIL) : FIX_MAX;
+        high = aumHigh.safeDiv(supply, CEIL);
         assert(low <= high); // not obviously true just by inspection
 
         return (low, high, 0);
--- a/contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
+++ b/contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
@@ -82,7 +82,7 @@ contract CurveStableMetapoolCollateral is CurveStableCollateral {
     {
         // {UoA/pairedTok}
         (uint192 lowPaired, uint192 highPaired) = tryPairedPrice();
-        require(lowPaired > 0 && highPaired < FIX_MAX, "invalid price");
+        require(lowPaired != 0 && highPaired != FIX_MAX, "invalid price");
 
         // {UoA}
         (uint192 aumLow, uint192 aumHigh) = _metapoolBalancesValue(lowPaired, highPaired);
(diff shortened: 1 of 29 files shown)


tbrent:
Still need to write regression tests for issue 21

tbrent:
@akshatmittal ready for review. I'm not the _most_ confident this implementation is exactly what we want; there are a lot of degrees of freedom here and lots of ways to do things differently. In short I decided we should make `(>0, FIX_MAX)` a valid price (previously anything with FIX_MAX high price _must_ have had 0 low price) and tried to maintain FIX_MAX propagation throughout all calculations of high prices. Confirmed that both regression tests I wrote fail on the starting commit.

akshatmittal:
Just as a reminder to myself (shared privately before): Decided to make the change for propagating `FIX_MAX`, decided _not_ to make the change where `(>0, FIX_MAX)` becomes valid.

tbrent:
@akshatmittal amended to reflect discussed approach

contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
```diff
@@ -172,13 +169,6 @@ contract CurveStableMetapoolCollateral is CurveStableCollateral {
...
         aumLow += lowPaired.mul(pairedBal, FLOOR);
-
-        // Add-in high part carefully
-        uint192 toAdd = highPaired.safeMul(pairedBal, CEIL);
-        if (aumHigh + uint256(toAdd) >= FIX_MAX) {
-            aumHigh = FIX_MAX;
-        } else {
-            aumHigh += toAdd;
-        }
+        aumHigh += highPaired.mul(pairedBal, CEIL);
```
akshatmittal, line 172: So, do we _want_ to revert if it overflows? I feel like the existing implementation here would be better for safeguarding the return. ~Or~ And should be using `safeMul`, I think.

akshatmittal: nvm

tbrent: Noting for others: we want to revert here because otherwise it's possible for low to be non-zero while high is FIX_MAX. Reverting here causes `tryPrice()` to revert which is what we want.

contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
```diff
@@ -124,8 +120,8 @@ contract CurveStableMetapoolCollateral is CurveStableCollateral {
     // Check for defaults outside the pool
     function _anyDepeggedOutsidePool() internal view virtual override returns (bool) {
         try this.tryPairedPrice() returns (uint192 low, uint192 high) {
-            // {UoA/tok} = {UoA/tok} + {UoA/tok}
-            uint192 mid = (low + high) / 2;
+            // D18{UoA/tok} = D18{UoA/tok} + D18{UoA/tok}
+            uint256 mid = (low + uint256(high)) / 2;
```
pmckelvy1, line 124: why do we need this?

pmckelvy1 (approved):

contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
tbrent: See here: https://github.com/code-423n4/2023-07-reserve-findings/issues/21

tbrent: Hmm I guess now that RTokenAsset always returns (0, FIX_MAX) instead of (>0, FIX_MAX), then this isn't necessary. Still, seems better to have protection here and soft default instead of having refresh() revert.

PR #918, opened 2023-08-25
facade support batch auctions

julianmrodri:
*Adapt `FacadeAct` to support different Trade Kinds in `nextRecollateralizationAuction`


tbrent:
LGTM!

PR #920, opened 2023-08-28
Clarify rewards handling on transfer

julianmrodri:
* Adds comment for `StaticATokenLM` to clarify that only those rewards previously collected are the ones that are updated on every `transfer`.  This addresses the following C4 issues:

C4 Issue #4 - https://github.com/code-423n4/2023-07-reserve-findings/issues/4
C4 issue #12 - https://github.com/code-423n4/2023-07-reserve-findings/issues/12

### DIFF
--- a/contracts/plugins/assets/aave/StaticATokenLM.sol
+++ b/contracts/plugins/assets/aave/StaticATokenLM.sol
@@ -366,6 +366,9 @@ contract StaticATokenLM is
 
     /**
      * @notice Updates rewards for senders and receiver in a transfer (not updating rewards for address(0))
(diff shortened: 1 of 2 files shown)


julianmrodri:
@pmckelvy1 @tbrent not sure how much detail or significance should we give to this issue in the comments. I added a minimum comment to the `_beforeTransfer` .. function but let me know if its better to add a paragraph at the top of the contract as well explaining this is more detail, or if this is OK.

tbrent:
I do think we should have a paragraph at the top

tbrent:
LGTM

PR #925, opened 2023-08-31
distribute token to buy in setDistribution

julianmrodri:
*Distributes RSR and RToken from the Traders before updating the Distributions
* Adjusts tests


tbrent:
Look perf, LGTM

PR #926, opened 2023-08-31
loss of pending rewards when claiming in StaticAToken

julianmrodri:
* add comments and test on loss of rewards when claiming with `forceUpdate=false`. 

* Addresses: C4  https://github.com/code-423n4/2023-07-reserve-findings/issues/10


pmckelvy1:
lgtm

PR #928, opened 2023-09-01
add tokens and feeds for base.

pmckelvy1:


common/configuration.ts
```diff
@@ -487,6 +487,28 @@ export const networkConfig: { [key: string]: INetworkConfig } = {
     },
     GNOSIS_EASY_AUCTION: '0xcdf32E323e69090eCA17adDeF058A6A921c3e75A', // mock
   },
+  '8453': {
+    name: 'base',
+    tokens: {
+      DAI: '0x50c5725949A6F0c72E6C4a641F24049A917DB0Cb',
+      USDC: '0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA',
```
akshatmittal, line 494: Just to be clear, this isn't the canonical USDC on Base, but the Coinbase briged eth-USDC. Need to be clear about it, and possibly have two different definitions for it. I expect some of DeFi to move to the actual USDC at some point (right?).

pmckelvy1: Good catch. This is USDbC, so I'll change the name. We might want to have both but im pretty sure the lending markets and dexes (and coinbase) are all leaning into usdbc.

pmckelvy1:
closing to base on 3.0.0

PR #933, opened 2023-09-07
advance oracles to prevent timeouts

tbrent:
Problem was that even though hardhat rolls back state between tests, it still progresses the chain time and blocks, resulting in our forked oracles eventually timing out. The solution is to refresh oracle values between each test. We were already doing this almost everywhere, just missing it in `deployCollateral()`. I also considered creating a new function in the suite `refreshOracles()` for each suite to implement, but this seemed slightly worse to me because the default fixture setup _was_ doing it correctly, it was just the `deployCollateral()` function that was not. 

I restored the previous ordering of the tests to prove this works. 

Also sped up the generic suite a bit by removing one unnecessary `resetFork()`.


pmckelvy1 (approved):
lgtm

PR #936, opened 2023-09-07
Fix maWBTC

tbrent:
`MorphoNonFiatCollateral` had improper oracle logic due to units being imprecise. It was also being instantiated with the same oracle address set to both the 2 addresses it needed. Wasn't caught in tests because of an improper `itIsPricedByPeg` setting. These issues existed in the deployment scripts too.


pmckelvy1 (approved):
lgtm

PR #938, opened 2023-09-08
3.0.0 executor

tbrent:
`FacadeWrite` was setting up executor incorrectly, so that anyone could execute. Bypasses same-era checks in `Governance`


julianmrodri:
Wouldn't this require a governance proposal (with its corresponding delay) to execute a proposal even if the time has passed and its ready to be executed?

tbrent:
No, shouldn't. Check out the `execute()` function on the governor. It's tested in `Governance.test.ts` as well

tbrent:
We were already executing proposals through the Governor in our tests

julianmrodri:
Oh ok. So anyone can execute on the Governor, and the Governor will execute on the Time Lock. You cannot execute directly on the Timelock anymore directly.

julianmrodri (approved):
LGTM but not sure why deployment scripts keep failing in CI

tbrent:
CI failing is present in `3.0.0-deploy`, so merging this since it is unrelated.

PR #939, opened 2023-09-08
Make FacadeAct.runRevenueOverview() work when toStart is empty

tbrent:
Not ready to merge; needs deploy

### DIFF
--- a/test/Facade.test.ts
+++ b/test/Facade.test.ts
@@ -1164,6 +1164,18 @@ describe('FacadeRead + FacadeAct contracts', () => {
           },
         ]
       )
+
+      // Nothing should be settleable
+      expect((await facade.auctionsSettleable(rsrTrader.address)).length).to.equal(0)
+
+      // Advance time till auction ended
+      await advanceBlocks(1 + auctionLength / 12)
+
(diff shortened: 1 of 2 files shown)


test/Facade.test.ts
```diff
@@ -1164,6 +1164,18 @@ describe('FacadeRead + FacadeAct contracts', () => {
...
       )
+
+      // Nothing should be settleable
+      expect((await facade.auctionsSettleable(rsrTrader.address)).length).to.equal(0)
+
+      // Advance time till auction ended
+      await advanceBlocks(1 + auctionLength / 12)
+
+      // Settleable now
+      expect((await facade.auctionsSettleable(rsrTrader.address)).length).to.equal(0)
```
julianmrodri, line 1175: Based on the comment I would expect a different thing here, seems its still not settleable?

tbrent: Very good question, investigating...

tbrent: Mmmm off-by-one-block read vs write difference. Advancing one further block solves the problem.

julianmrodri (approved):
LGTM

PR #948, opened 2023-09-15
use infura in deployment scripts

julianmrodri:
* Use Infura instead of Alchemy for the deployment scripts. This is required due to a bug in Alchemy which needs to be fixed based on feedback from both OZ and Hardhat.
* Allows to unify into single `mainnet.json` file for the OZ plugin
* Does not replace all uses of Alchemy because the public end point seems to not be able to handle the workload of our tests. So we still need Alchemy.


pmckelvy1 (approved):
lgtm

PR #951, opened 2023-09-21
feat: add docgen Github Action

larrythecucumber321:
Uses Foundry's [docgen](https://github.com/foundry-rs/foundry/tree/master/crates/doc) utility to generate dev docs, which are then pushed to Vercel at https://developer.reserve.org/. The GA triggers on pushes to master.

(note, my free-tier Vercel has a rate-limit of one run per 24 hrs, so failures will not be uncommon)

image


pmckelvy1:
> (note, my free-tier Vercel has a rate-limit of one run per 24 hrs, so failures will not be uncommon)
lets's just put it on whatever deployment pipline kevin is using?

pmckelvy1:
ideally we could generate these docs and leave out `p0`. there doest seem to be a way to exclude files in the current `forge doc` command though

PR #953, opened 2023-09-25
Plugin stargate fixes and changes

julianmrodri:
* Deprecate `StargatePoolEthCollateral.sol`
* Fix bug in rewards for `StargatePoolFiatCollateral.sol` and inherit from `AppreciatingFiatCollateral`
* Adapt deployment scripts


scripts/deployment/phase2-assets/collaterals/deploy_stargate_usdt_collateral.ts
```diff
@@ -70,17 +70,20 @@ async function main() {
...
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDT!,
+      oracleError: fp('0.0025').toString(), // 0.25%,
+      erc20: erc20.address,
+      maxTradeVolume: fp('1e6').toString(), // $1m,
+      oracleTimeout: oracleTimeout(chainId, '86400').toString(), // 24h hr,
+      targetName: hre.ethers.utils.formatBytes32String('USD'),
+      defaultThreshold: fp('0.0125').toString(), // 1.25%
+      delayUntilDefault: bn('86400').toString(), // 24h
+    },
+    fp('0') // no revenueHiding
```
tbrent, line 85: @akshatmittal how did you conclude [it needs no revenue hiding?](https://github.com/reserve-protocol/protocol/pull/863#issuecomment-1622668093) 

did you check their math and ensure the roundings are correct? you may have, but I just wanted to check

tbrent (commented):
LGTM other than question about revenue hiding

julianmrodri:
Also we need to pay special attention to the `_claimRewards()` function in the wrapper which was simplified based on Trust input, but we need more confirmation this is ok.

scripts/deployment/phase2-assets/collaterals/deploy_stargate_usdt_collateral.ts
akshatmittal: @tbrent Initially I thought it will always go up, but I later learned that it can indeed go down with liquidity changes. So we do need to introduce some revenue hiding, need to figure out how much.

tbrent: @julianmrodri let's add `fp('1e-6')` revenue hiding to stargate across the board

tbrent (approved):
Revenue hiding changes LGTM

PR #954, opened 2023-09-25
Aave v3 review

tbrent:
Added some comments for documenting where we got the wrapper from and the level of auditing/formal verification that occurred. 

My other 2 observations:
- We don't have deployment/verification scripts, which we need before calling the plugin complete. It's possible for the deployment script to be different than the testing arguments, so this needs its own review.
- There's no reward claiming tests in the test suite. I tried to add a test thinking `stkAAVE` would be one of the (multiple) reward tokens that might be earned, but looks like this isn't the case. Anyway, it would be good to actually have a test-case that earns rewards; we only have one that checks that it does not revert. 

Follow-up: Created deployment/verification scripts. Tried to test rewards for an hour but it's really difficult, so just satisfied myself by carefully reading the `claimRewards()` function we added; the rest of the code is formally verified.


tbrent:
I do believe the .sol plugin and the wrapper are correct, though.

pmckelvy1:
lgtm

PR #955, opened 2023-09-26
Base lsds (just cbETH for now)

pmckelvy1:


common/configuration.ts
```diff
@@ -474,6 +481,34 @@ export const networkConfig: { [key: string]: INetworkConfig } = {
...
+    name: 'base',
+    tokens: {
+      DAI: '0x50c5725949A6F0c72E6C4a641F24049A917DB0Cb',
+      USDbC: '0xd9aAEc86B65D86f6A7B5B1b0c42FFA531710b6CA',
+      RSR: '0xaB36452DbAC151bE02b16Ca17d8919826072f64a',
+      COMP: '0x9e1028F5F1D5eDE59748FFceE5532509976840E0',
+      WETH: '0x4200000000000000000000000000000000000006',
+      cbETH: '0x2ae3f1ec7f1f5012cfeab0185bfc7aa3cf0dec22',
+      cUSDbCv3: '0x9c4ec768c28520B50860ea7a15bd7213a9fF58bf',
+      aUSDbCv3: '0x0a1d576f3eFeF75b330424287a95A366e8281D54',
```
julianmrodri, line 494: Symbols differ a bit for these tokens in Base, but I guess this was done on purpose to simplify and not depend on the network?

julianmrodri (changes requested):
Looks good overall left some comments

common/configuration.ts
pmckelvy1: USDbC is whats used in Aave V3 and Compound V3 on base. it's different from USDC (USDbC is bridged)

julianmrodri (approved):
lgtm if tests pass

PR #958, opened 2023-10-03
swap _beforeWithdraw if statements for efficiency.

pmckelvy1:


pmckelvy1:
just swapped the logic around. this should now be more efficient, especially if allocPoint gets set to 0 (it will only call `emergencyWithdraw` once)

julianmrodri (approved):
LGTM

tbrent:
Took me a second to understand why this is okay, but LGTM now

PR #959, opened 2023-10-03
max custom redeem algo

tbrent:
This touched a lot of files because of breaking apart `FacadeRead` into two facades, but the meaningful part of the diff is just the new `FacadeMint` contract and the new test: `Should calculate customRedeem portions correctly`. 

Open to better names for the `FacadeMint`. Issue seemed like a worse word to use as a suffix, and we've used suffixes elsewhere.


tbrent:
After approval I'll deploy

contracts/facade/FacadeMint.sol
```diff
@@ -0,0 +1,288 @@
...
+
+        // Populate `_erc20Bals` with the registered erc20s and their initial balances
+        {
+            IERC20[] memory erc20s = rToken.main().assetRegistry().erc20s();
+            for (uint256 i = 0; i < erc20s.length; i++) {
+                _erc20Bals.set(address(erc20s[i]), erc20s[i].balanceOf(bm));
+            }
+        }
+
+        // Walk backwards from current nonce, deducting from basketsNeeded greedily
```
pmckelvy1, line 210: comment is wrong. seems we're walking forward instead

test/Facade.test.ts
```diff
@@ -219,819 +203,917 @@ describe('FacadeRead + FacadeAct contracts', () => {
...
+        await main.connect(owner).freezeShort()
+        await expect(facadeMint.callStatic.redeem(rToken.address, issueAmount)).to.be.revertedWith(
+          'frozen'
+        )
+      })
 
-      // Check values - No supply, returns 0
-      expect(backing).to.equal(0)
-      expect(overCollateralization).to.equal(0)
+      itP1('Should calculate customRedeem portions correctly', async () => {
```
pmckelvy1, line 377: seems like we should also test at least 1 or 2 scenarios that are a little more complex than this?

pmckelvy1 (changes requested):

tbrent:
@akshatmittal @pmckelvy1 this algo turns out to be insufficient. Going either solely forwards or backwards does not work all the time. I think the fastest simple algo is `O(n^2)`: you have to pass over each basket nonce and choose the one with the max number of baskets given current available balances, and you have to do this for every single choice of next basket nonce to use. 

I could imagine discovering there's a `O(n*log(n)) algo somewhere out there if you do a sort on some clever computed metric, but a simple `O(n)` algo does not exist pretty sure.

tbrent:
One not so crazy option is to limit the search to the past `n` basket nonces, where `n` is something like 3 or 5. Or at least, we'd take this as an argument and leave it to the off-chain component to decide which `n` to use.

tbrent:
This is absolutely a sign that this problem is best-solved fully from off-chain, though.

akshatmittal:
I think so too, this should be off-chain ideally. Seems my doubts were reasonable, thanks for looking into it!

tbrent:
I went to start writing the `O(n^2)` solution and it very quickly got past the stack limit. We should close out this PR and not use it. This also means we don't have to split up `FacadeRead` just yet.

PR #960, opened 2023-10-04
get gas measurements, fix ftoken/ctoken vault gas.

pmckelvy1:


tbrent (approved):
LGTM

I realized after reviewing this I could probably get us ERC20 transfer measurements pretty quickly so I wrote #962. Ideally we'd measure all 3-of (i) deposit; (ii) transfer; (iii) redeem; but just the transfer measurements should be useful to start. Convex transfers pretty high.

tbrent (changes requested):
Looks good -- I checked the output of each function for each collateral address, and confirmed the ERC20s are correct. 

Only thing to add is we need a change to `scripts/deploy.ts` the new `deploy_compound_v2_collateral.ts` script

(I see we ran into the `refresh()` out of gas error even with 2x gas multiplier on)

PR #961, opened 2023-10-05
rollback unwrapping changes.

pmckelvy1:
i don't think we necessarily want to do this. just made this pr so we could discuss based on the harder gas numbers in the target branch `fix-ftoken`.

unwrapped ctoken/ftoken transfer: `110849`
wrapped ctoken/ftoken transfer: `175997`
+59% increase

issue, RToken, single collateral basket unwrapped ctoken/ftoken: `677455`
issue, RToken, single collateral basket wrapped ctoken/ftoken: `753143`
+11% increase
(11% will go down as more tokens that are not ctoken/ftoken are added to the basket. ie. eUSD difference will be closer to 6%)


tbrent:
I must be missing something: the numbers show it's more costly to transfer wrapped tokens than unwrapped, no?

PR #962, opened 2023-10-05
Erc20 wrapper tests everywhere

tbrent:
Quick PR to surface our current ERC20 transfer costs.

Any gas tests for deposit/redeem flows would have to go in the specific wrapper tests and not the collateral suites, but I wanted to get something quickly that could show us ERC20 transfer costs without much effort. The transfer side is easily measurable so I added that.

I'm surprised at how high the Convex wrapper numbers are.


akshatmittal:
This is good to have.

Yeah, Convex is quite high, and Morpho would also have been high if they weren't themselves a wrapper to Aave.

PR #965, opened 2023-10-06
aavev3 and compoundv3 base tests

julianmrodri:
* Sets up for AaveV3 and Compound V3 to run on Base as well.
* Only had to skip one test in CToken Wrapper tests, which required rewards to be enable and could not find a block to use. Its not a critical test and its already verified on Mainnet.


test/plugins/individual-collateral/aave-v3/constants.ts
```diff
@@ -1,2 +1,32 @@
-export const AAVE_V3_USDC_POOL = '0x87870Bca3F3fD6335C3F4ce8392D69350B4fA4E2'
-export const AAVE_V3_INCENTIVES_CONTROLLER = '0x8164Cc65827dcFe994AB23944CBC90e0aa80bFcb'
+import { networkConfig } from '../../../../common/configuration'
+import { useEnv } from '#/utils/env'
+
+const forkNetwork = useEnv('FORK_NETWORK') ?? 'mainnet'
```
pmckelvy1, line 4: either use this in the test file or delete it here

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
```diff
@@ -639,15 +641,17 @@ describeFork('Wrapped CUSDCv3', () => {
...
+      expect(expectedBalanceBob).to.be.closeTo(
+        await compToken.balanceOf(don.address),
+        expectedBalanceBob.mul(5).div(1000)
+      ) // within 0.5%
     })
 
     // In this forked block, rewards accrual is not yet enabled in Comet
-    it('claims no rewards when rewards accrual is not enabled', async () => {
+    // Only applies to Mainnet forks (L1)
+    itL1('claims no rewards when rewards accrual is not enabled', async () => {
```
pmckelvy1, line 653: why don't we want to test this on base?

test/plugins/individual-collateral/compoundv3/constants.ts
```diff
@@ -1,17 +1,39 @@
 import { bn, fp } from '../../../../common/numbers'
 import { networkConfig } from '../../../../common/configuration'
+import { useEnv } from '#/utils/env'
+
+const forkNetwork = useEnv('FORK_NETWORK') ?? 'mainnet'
```
pmckelvy1, line 5: either import this in the test.ts file or delete it

pmckelvy1 (changes requested):

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
```diff
@@ -639,15 +641,17 @@ describeFork('Wrapped CUSDCv3', () => {
       await advanceBlocks(1)
 
       expect(await compToken.balanceOf(bob.address)).to.be.greaterThan(0)
-      expect(await compToken.balanceOf(bob.address)).to.equal(
-        await compToken.balanceOf(don.address)
-      )
+      const expectedBalanceBob = await compToken.balanceOf(bob.address)
```
pmckelvy1, line 644: why do we have to loosen this check?

pmckelvy1 (changes requested):

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
julianmrodri: Requires a fork block by default with rewards disabled. Could not find in Base. If there is one we can try.

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
julianmrodri: There is a small difference i the numbers. Maybe due to logic based on blocks/network. It was 1 atto token in one case and less than 0.5% in another (maybe less even).

julianmrodri:
Fixed the `forkNetwork` duplication. Added explanation for the other changes.

pmckelvy1:
lgtm

PR #966, opened 2023-10-09
Trust M-01 & M-02: RevenueTrader.settleTrade()

julianmrodri:
* Do not attempt to distribute on `RevTrader.settleTrade()` if distribution is zero
* Do not attempt to distribute on `RevTrader.settleTrade()` if trading paused or frozen
* Do not start new trades in `RevTrader.manageTokens()` if distribution is zero

### DIFF
--- a/contracts/p0/RevenueTrader.sol
+++ b/contracts/p0/RevenueTrader.sol
@@ -32,14 +32,12 @@ contract RevenueTraderP0 is TradingP0, IRevenueTrader {
     /// @param sell The sell token in the trade
     /// @return trade The ITrade contract settled
     /// @custom:interaction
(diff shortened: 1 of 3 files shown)


tbrent (commented):
Noting that I merged this pre-emptively by fat-fingering the button when I thought I was looking at #968 

Thought, though: it seems gas inefficient to do it this way. That is, checking to see whether we expect `_distributeTokenToBuy` will succeed in advance is going to be more gas wasteful than simply trying to distribute the token to buy. What about instead of this change we simply do a try-catch on `distributeTokenToBuy()`? This avoids doing the lookup in the distribution table twice.

PR #968, opened 2023-10-09
Trust M-11: Fix storage slots

julianmrodri:
Fix storage slots in Broker. Confirmed and fixed using regression testing with the `validate-upgrade` task.


tbrent:
LGTM

PR #969, opened 2023-10-10
Trust M-01 & M-02: RevenueTrader.settleTrade()

julianmrodri:
* Do not attempt to distribute on RevTrader.settleTrade() if distribution is zero
* Do not attempt to distribute on RevTrader.settleTrade() if trading paused or frozen
* Do not start new trades in RevTrader.manageTokens() if distribution is zero


tbrent:
I've added OOG checks to the catch statements to prevent a trade settler from skipping `_distributeTokenToBuy()`

Tests are thorough, LGTM

PR #971, opened 2023-10-10
Max approval erc20s

tbrent:


contracts/libraries/Allowance.sol
```diff
@@ -0,0 +1,32 @@
...
+    ///   3. Falls back to setting a maximum allowance, if (2) fails
+    /// Context: Some new-age ERC20s think it's a good idea to revert for allowances
+    /// that are > 0 but < type(uint256).max.
+    function safeApproveFallbackToMax(
+        IERC20 token,
+        address spender,
+        uint256 value
+    ) internal {
+        // 1. Set initial allowance to 0
+        require(token.approve(spender, 0), "failed to set 0 allowance");
```
akshatmittal, line 19: Some tokens do not return true/false. This should either be SafeERC20 or do a proper allowance return check flow.

contracts/p1/mixins/Trading.sol
```diff
@@ -120,8 +121,12 @@ abstract contract TradingP1 is Multicall, ComponentP1, ReentrancyGuardUpgradeabl
...
         assert(address(trades[sell]) == address(0));
 
-        IERC20Upgradeable(address(sell)).safeApprove(address(broker), 0);
-        IERC20Upgradeable(address(sell)).safeApprove(address(broker), req.sellAmount);
+        // Set allowance via custom approval -- first sets allowance to 0, then sets allowance
+        // to either the requested amount or the maximum possible amount, if that fails.
+        //
+        // Context: wcUSDCv3 has a non-standard approve() function that reverts if the approve
+        // amount is > 0 and < type(uint256).max.
+        AllowanceLib.safeApproveFallbackToMax(req.sell.erc20(), address(broker), req.sellAmount);
```
akshatmittal, line 129: If we are specifically making the changes for Trading stuff, should we consider doing the same in other places?

contracts/libraries/Allowance.sol
pmckelvy1: sup let's use safeApprove like we do in other places

tbrent: Yeah. Also, just changed the implementation, take a 2nd look.

contracts/p1/mixins/Trading.sol
tbrent: We should have hit all the places that we do an approval for a non-type(uint256).max amount.

pmckelvy1: the only other place would be RevenueTrader._distributeTokenToBuy(), which should only be dealing with an RToken or RSR.

maybe its worth adding a comment about assuming `tokenToBuy` is RToken or RSR?

contracts/libraries/Allowance.sol
```diff
@@ -0,0 +1,39 @@
...
+        bool success; // bool success = false;
+        try token.approve(spender, value) returns (bool) {
+            success = token.allowance(address(this), spender) == value;
+        } catch {}
+
+        // 3. Fall-back to setting a maximum allowance
+        if (!success) {
+            token.approve(spender, type(uint256).max);
+            require(
+                token.allowance(address(this), spender) == type(uint256).max,
```
akshatmittal, line 34: Remind me, wasn't there a token specifically which doesn't use uint256 for the approval max?

I don't think we need this require statement here.

tbrent: Actually, went with keeping the require but changing it to `>= value`

tbrent:
LGTM

PR #972, opened 2023-10-11
TRUST M-06: Keep buffer when minting rtoken revenue

julianmrodri:
Maintan always the correct backing buffer level when minting RToken revenue


tbrent:
LGTM

PR #973, opened 2023-10-11
Trust M-03: track balances out on trade for RTokenAsset.price()

tbrent:
This PR adds the tracking of balances to the BackingManager, for use in `RecollateralizationLibP1.basketRange()` when used by `RTokenAsset` for pricing. 

I put the `bals` array in the TradingContext in order to avoid re-computing available balances throughout `basketRange()` and `nextTradePair()`. I also moved the calculation of the TradingContext struct into a single view function on the BackingManager in order to avoid duplicating code in RTokenAsset. I think it's also more gas-efficient this way. 

I also had to add a new function to `GnosisTrade` in order to efficiently access the sell amount in `{tok}` units without adding specific logic that checks the trade kind, which seemed worse. This increases gas costs of `GnosisTrade` slightly, which should be fine given it will be the minority trading method used.


tbrent:
Wow all previous tests pass, you love to see it

pmckelvy1 (changes requested):

contracts/p1/mixins/RecollateralizationLib.sol
```diff
@@ -329,26 +282,33 @@ library RecollateralizationLibP1 {
         Registry memory reg,
         BasketRange memory range
     ) private view returns (TradeInfo memory trade) {
+        // assert(tradesOpen == 0); // all balances in ctx are guaranteed to be available
```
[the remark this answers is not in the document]
tbrent: I'm down to actually assert it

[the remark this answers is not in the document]
tbrent: Oh right, it's because `tradesOpen` is in `BackingManager` and we're in the recollat library here. There's a require in `BackingManager.rebalance()` that checks this.

contracts/p1/BackingManager.sol
```diff
@@ -149,22 +153,26 @@ contract BackingManagerP1 is TradingP1, IBackingManager {
...
+        ) = RecollateralizationLibP1.prepareRecollateralizationTrade(ctx, reg);
 
         if (doTrade) {
+            IERC20 sellERC20 = req.sell.erc20();
+
             // Seize RSR if needed
-            if (req.sell.erc20() == rsr) {
+            if (sellERC20 == rsr) {
                 uint256 bal = req.sell.erc20().balanceOf(address(this));
                 if (req.sellAmount > bal) stRSR.seizeRSR(req.sellAmount - bal);
```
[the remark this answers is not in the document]
tbrent: I think it's fine? Agree that `tokensOut[rsr]` contains balances of RSR from both the BackingManager and StRSR addresses. There are no balances out on trade because `tradesOpen` is 0.

test/plugins/Asset.test.ts
```diff
@@ -505,6 +515,94 @@ describe('Assets contracts #fast', () => {
...
 
+    it('Should handle tokens being out on trade for RTokenAsset', async () => {
+      // Summary:
+      // - Run a dutch auction that does not fill
+      // - Run a batch auction that fills for partial volume
+      // - Run a dutch auction that fills for full volume
+
+      const low0 = fp('0.99')
+      const low1 = bn('975344098811881188') // after a 50% basket change
+      const low2 = bn('975343128415841584') // after batch auction at half volume
```
pmckelvy1, line 526: low price falls? isn't it an invariant that this never happens?

pmckelvy1 (changes requested):

test/plugins/Asset.test.ts
tbrent: Every increase of `basketsHeld.bottom` (ie a successful round of purchasing full baskets) raises `RTokenAsset.price().low`, but there are small drops as the full baskets are purchased.

tbrent: I've changed a comment in commit aecbf64c16ba2191469d6eb152da1b6d33d12a26 to reflect this. Previous comment was actually technically wrong.

pmckelvy1: ok i've looked deeper at the code and this makes sense to me now, thanks for the explanation

tbrent:
Going to leave this open for a few days and merge it later once (i) we've marinated on the `basketRange().low` decrease; and (ii) all the other pending 3.1.0 branches go in

PR #976, opened 2023-10-12
TRUST Recommendation 1: reward and underlying checks in RewardERC20 wrappers

julianmrodri:
* adds check to ensure reward and underlying tokens are not the same for ERC20 wrappers.


tbrent:
LGTM

PR #981, opened 2023-10-17
TRUST L-7: Restriction on reportViolation

julianmrodri:
* Allow to reportViolation() when paused or frozen
* Replace test


test/Revenues.test.ts
```diff
@@ -2632,6 +2632,122 @@ describe(`Revenues - P${IMPLEMENTATION}`, () => {
...
+            name: 'TradeStarted',
+            emitted: false,
+          },
+          {
+            contract: rTokenTrader,
+            name: 'TradeStarted',
+            emitted: false,
+          },
+        ])
+
```
tbrent, line 2740: This is a nit, but let's add the explicit `expect(await broker.batchTradeDisabled()).to.equal(true)` here, even though we have the event.

tbrent (commented):
LGTM, just left a tiny nit on the tests. 

Also, so we don't drop this part later: an you merge #964 into this branch and adjust `docs/pause-freeze-states.md` to reflect the new absence of the modifier? There's a table in that file of when different functions are enabled/disabled as a function of paused/frozen states.

PR #982, opened 2023-10-17
Trst m 14

tbrent:
- Move oracle timeout buffer into constructor of contract and out of deployment scripts
- Change buffer from 60s to 300s
- Kruft cleanup: delete curve/convex volatile collateral scripts

A lot of the test changes come from the fact that tests need to both be able to instantiate a new asset (use pre-buffer oracleTimeout) and also need to be able to fast-forward to the end of the oracleTimeout to cause an oracle to become stale. It would be nice to just always read this value off of an asset, but often we're fast forwarding outside of the context of any single asset.


test/Recollateralization.test.ts
```diff
@@ -3248,7 +3249,6 @@ describe(`Recollateralization - P${IMPLEMENTATION}`, () => {
                 collateral1.address,
                 collateral0.address,
                 issueAmount,
-                config.minTradeVolume,
```
pmckelvy1, line 3251: kind of worried about this change since the function is not edited in this pr, meaning there might have been something wrong with this test?

pmckelvy1 (changes requested):

pmckelvy1:
lgtm

test/Recollateralization.test.ts
tbrent: lol looks like `config.minTradeVolume` was set to `fp('1e-2')` and `config.maxTradeSlippage` set to `fp('0.01')`, so we're all good here.

tbrent: I also confirmed by hand this test was running properly

PR #984, opened 2023-10-18
TRUST M-4: Call reward accounting functions

julianmrodri:
Call reward accounting functions after `forwardRevenue()` and `distribute()`. Add tests.


tbrent (commented):
Looks like there's a failing test that might be real, but implementation looks good to me. I don't see a more gas-efficient way to do it.

PR #986, opened 2023-10-20
TRUST QA-5/10/11: Optimizations on BasketHandler

julianmrodri:
* Replaces `continue` with `break` to optimize gas
* Uses `lastStatus` to optimize gas
* Tracks status when basket is disabled (keeps state consistent)

### DIFF
--- a/contracts/p0/BasketHandler.sol
+++ b/contracts/p0/BasketHandler.sol
@@ -183,6 +183,8 @@ contract BasketHandlerP0 is ComponentP0, IBasketHandler {
         }
         emit BasketSet(nonce, basket.erc20s, refAmts, true);
         disabled = true;
+
+        trackStatus();
     }
 
     /// Switch the basket, only callable directly by governance or after a default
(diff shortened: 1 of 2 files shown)


contracts/p0/BasketHandler.sol
```diff
@@ -458,7 +460,7 @@ contract BasketHandlerP0 is ComponentP0, IBasketHandler {
                 for (uint256 k = 0; k < len; ++k) {
                     if (b.erc20s[j] == erc20sAll[k]) {
                         erc20Index = k;
-                        continue;
+                        break;
```
tbrent, line 463: Since this is P0 we should not do the gas optimization here

tbrent (commented):
I believe `BasketHandler.refreshBasket()` can use stored lastStatus is _not_ a valid recommendation from Trust. There are 2 things wrong with it:
1. Simply time passing can cause a collateral to become DISABLED, based on its `whenDefault`
2. Collaterals can be interacted with directly without going through the assetRegistry

So we should not do QA10

PR #987, opened 2023-10-20
TRUST QA-4: Smoother exposedRefPrice in case of default

julianmrodri:
Sets `exposedRefPrice` to `underlyingRefPerTok` in case of default


tbrent:
LGTM

tbrent:
Actually, adding some tests...

PR #988, opened 2023-10-20
TRUST QA-6/7: Optimizations in Distributor

julianmrodri:
* Prevent zero transfers and executing logic if nothing to distribute (due to rounding or no balance)
* Simplify struct


tbrent:
LGTM!

PR #990, opened 2023-10-24
TRUST QA-21: minimum voting delay

julianmrodri:
* Enforce Minimum voting delay of 1 day (dependes on Network Config)
* Adds test

### DIFF
--- a/contracts/plugins/governance/Governance.sol
+++ b/contracts/plugins/governance/Governance.sol
@@ -8,6 +8,7 @@ import "@openzeppelin/contracts/governance/extensions/GovernorTimelockControl.so
 import "@openzeppelin/contracts/governance/extensions/GovernorVotes.sol";
 import "@openzeppelin/contracts/governance/extensions/GovernorVotesQuorumFraction.sol";
 import "../../interfaces/IStRSRVotes.sol";
+import "../../libraries/NetworkConfigLib.sol";
 
 /*
  * @title Governance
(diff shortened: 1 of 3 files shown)


contracts/plugins/governance/Governance.sol
```diff
@@ -44,7 +47,12 @@ contract Governance is
         GovernorVotes(IVotes(address(token_)))
         GovernorVotesQuorumFraction(quorumPercent)
         GovernorTimelockControl(timelock_)
-    {}
+    {
+        MIN_VOTING_DELAY =
+            (86400 + NetworkConfigLib.blocktime() - 1) /
```
tbrent, line 52: Let's declare the 86400 figure as a constant just outside the contract

contracts/plugins/governance/Governance.sol
```diff
@@ -30,7 +31,9 @@ contract Governance is
     // 100%
     uint256 public constant ONE_HUNDRED_PERCENT = 1e8; // {micro %}
 
-    // solhint-disable no-empty-blocks
+    // solhint-disable-next-line var-name-mixedcase
+    uint256 public immutable MIN_VOTING_DELAY; // in blocks, based on network
```
tbrent, line 35: let's reference here in the comment that it should be equal to `ONE_DAY`, or whatever we end up calling the 86400 figure as a constant

tbrent (commented):
Tests look good

contracts/plugins/governance/Governance.sol
julianmrodri: will do the changes and push

tbrent: also can use our unit notation `{blocknumber}` here

PR #994, opened 2023-10-26
TRUST M-8: Reward distribution in `RewardableERC20`

julianmrodri:
*Fixes incorrect reward distribution in RewardableERC20 by adding a decimals buffer = 9.

*Adds regression testing with the example from Trust with the exact values to confirm its distributing correctly

*Adapts tests, now shares and balance are in different units. So we need to multiply or divide by the "Shares Factor" (`bn(10).pow(SHARES_DECIMALS=9)`) to get the previous comparisons to work. This is the only change accross the file.

### DIFF
--- a/contracts/plugins/assets/erc20/RewardableERC20.sol
+++ b/contracts/plugins/assets/erc20/RewardableERC20.sol
@@ -7,6 +7,9 @@ import "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import "../../../interfaces/IRewardable.sol";
 
+uint256 constant SHARES_BUFFER_DECIMALS = 9; // to prevent reward rounding issues
+
+/*
 /**
  * @title RewardableERC20
  * @notice An abstract class that can be extended to create rewardable wrapper.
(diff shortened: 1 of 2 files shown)


tbrent (commented):
So, I understand why making `one` 1e9 bigger fixes `_claimAndSyncRewards()`, but doesn't it break `_syncAccount()`? 

L:69 - `_accumulatedRewards += (delta * shares) / one;`

The `shares` is no bigger than it was, while the denominator is 1e9 bigger now. Doesn't this make `accumulatedRewards[account]` too small?

julianmrodri:
> The `shares` is no bigger than it was, while the denominator is 1e9 bigger now. Doesn't this make `accumulatedRewards[account]` too small?

Isnt the `delta` used in the numerator also increased by 1e9?  That value comes from `rewardsPerShare` which is multiplied by `one` when calculated as well. But I may be wrong.

pmckelvy1:
> > The `shares` is no bigger than it was, while the denominator is 1e9 bigger now. Doesn't this make `accumulatedRewards[account]` too small?
> 
> Isnt the `delta` used in the numerator also increased by 1e9? That value comes from `rewardsPerShare` which is multiplied by `one` when calculated as well. But I may be wrong.

i think this is correct.  i also think its worth moving the 2 extra `describe` blocks for "Special Case..." tests into in the `runTests` function so that theyre covered properly

tbrent:
I talked to Julian and confirmed it is working correctly, but I've asked him to go through and adjust all the unit comments to be correct so this is obvious

julianmrodri:
Implemented comments in units and moved tests to the main `runTests` function as recommended. I left the isolated special cases as well because they are more precise and serve as a reminder o regression tests identified by auditors.

tbrent:
LGTM

PR #995, opened 2023-10-27
TRUST QA-20:  Fix morpho DAI oracle timeout mainnet

julianmrodri:
* Fixes the DAI oracle timeout for the Morpho plugin

REQUIRES REDEPLOYMENT OF MORPHO ADAI TO MAINNET


tbrent:
LGTM, and confirmed that the CHANGELOG is sufficiently up to date so this does not get lost.

PR #997, opened 2023-10-30
TRUST L-5: add comment for non supported metapools

julianmrodri:
* Add comment for non-supported metapools that have a separate contract for the LP Token.


tbrent:
LGTM

PR #1002, opened 2023-10-31
Audit/trust m12

jankjr:
- [x] implemented suggeted fix
- [x] implemented a test that makes sure withdrawals are larger than deposits if time passes, disabled for stETH as it does not seem to appreciate

### DIFF
--- a/test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
+++ b/test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
@@ -9,6 +9,8 @@ import { loadFixture } from '@nomicfoundation/hardhat-network-helpers'
 import { bn } from '#/common/numbers'
 import { getResetFork } from '../helpers'
 import { FORK_BLOCK } from './constants'
+import { advanceBlocks, advanceTime } from '#/utils/time'
+import * as hre from 'hardhat'
 
 type ITokenSymbol = keyof ITokens
 const networkConfigToUse = networkConfig[31337]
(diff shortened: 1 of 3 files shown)


test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
```diff
@@ -29,12 +31,12 @@ const mkTestCase = <T extends ITokenSymbol>(
 })
 
 const TOKENS_TO_TEST = [
-  mkTestCase('USDC', '1000.0', '1'),
-  mkTestCase('USDT', '1000.0', '1'),
-  mkTestCase('DAI', '1000.0', '1'),
-  mkTestCase('WETH', '200.0', '1'),
+  // mkTestCase('USDC', '1000.0', '1'),
```
tbrent, line 34: I don't think you meant to comment these out?

test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
```diff
@@ -233,6 +235,28 @@ const execTestForToken = ({
       await closeTo(methods.balanceUnderlying(alice), fraction(100))
     })
 
+    it('Withdrawals larger than deposits', async () => {
```
tbrent, line 238: I just ran this test with the `updateIndexes()` calls commented out, and it passed.

tbrent (changes requested):

test/plugins/individual-collateral/morpho-aave/MorphoAaveV2TokenisedDeposit.test.ts
jankjr: Hm, so the change makes no difference?

tbrent:
Turns out this is unneeded, Trust has confirmed

PR #1003, opened 2023-10-31
TRUST H-3:  calling incorrect hasPermission() function

julianmrodri:
* Fixes  CusdcV3Wrapper._deposit() to call the correct hasPermission() function
* Now permission has to be done on the actual token that is going to be spent (`cuscdv3`) and not on the wrapper for depositing on behalf of someone else
* Adapts tests


tbrent:
LGTM, can go in after CI

PR #1004, opened 2023-10-31
add lotPrice() back everywhere (unused) for backwards compatibility

tbrent:
I realized while writing down 3.1.0 upgrade instructions that we can't pull `lotPrice()` off the assets + basketHandler if this is going to be a minor release. I've added it back everywhere with `lotPrice()` just returning the result of `price()`.

### DIFF
--- a/contracts/interfaces/IAsset.sol
+++ b/contracts/interfaces/IAsset.sol
@@ -32,6 +32,13 @@ interface IAsset is IRewardable {
     /// @return high {UoA/tok} The upper end of the price estimate
     function price() external view returns (uint192 low, uint192 high);
 
+    /// Should not revert
+    /// lotLow should be nonzero when the asset might be worth selling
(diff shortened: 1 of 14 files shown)


pmckelvy1:
worth adding back all the relevant tests?

tbrent:
> worth adding back all the relevant tests?

Well the behavior changed, the old tests would fail now. I do think it's worth adding new tests though just to make sure `lotPrice()` is implemented and that it returns the same thing as `price()`, will do.

tbrent:
I've added tests to check that (i) `lotPrice()` is on all asset interfaces; and (ii) `lotPrice()` returns the same thing as `price()`

pmckelvy1:
let's add them for RTokenAsset and BasketHandler as well

tbrent:
> let's add them for RTokenAsset and BasketHandler as well

Good catch, added

PR #1006, opened 2023-11-02
TRUST M-9: pegprice in `frxETH`

julianmrodri:
* Adds comments clarifying that the plugin is not ready to be used in Production because no oracle is available yet

### DIFF
--- a/contracts/plugins/assets/frax-eth/SFraxEthCollateral.sol
+++ b/contracts/plugins/assets/frax-eth/SFraxEthCollateral.sol
@@ -57,7 +57,7 @@ contract SFraxEthCollateral is AppreciatingFiatCollateral {
         high = p + err;
         // assert(low <= high); obviously true just by inspection
 
-        // Note: Currently not checking for depegs between `frxETH` and `ETH`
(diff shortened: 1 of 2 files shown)


contracts/plugins/assets/frax-eth/SFraxEthCollateral.sol
```diff
@@ -51,6 +57,8 @@ contract SFraxEthCollateral is AppreciatingFiatCollateral {
         high = p + err;
         // assert(low <= high); obviously true just by inspection
 
+        // Note: Currently not checking for depegs between `frxETH` and `ETH`
```
tbrent, line 60: Can we make this a `TODO`? Can keep the rest of the text, I just like having the string to search for

julianmrodri: Yeah I first did a TODO but then realized we dont have ANY in any of our contracts. But I can do it.

julianmrodri: Done

PR #1007, opened 2023-11-02
TRUST H-4: Single reward token

julianmrodri:
* Adds comments to `CurveGaugeWrapper` and `RewardableERC20` that only one single reward token is supported


tbrent (approved):
LGTM

PR #1016, opened 2023-11-28
Wcusdcv3 rewards fix

tbrent:
- Add regression test for large COMP reward amounts; confirmed fails on previous implementation
- Implement fix to avoid overflow. I considered changing all uint64 to uint256, but this would break the re-use of comet functions in `CometHelper`. It would actually be a much larger change to do this, and I don't think it's worth the cost. Just changing the storage type for `baseTrackingAccrued` is I think the minimal fix that is also slightly more robust than just adding a `uint256` cast to the broken multiplication in L:199. 

Needs review from everyone

### DIFF
--- a/contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
+++ b/contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
@@ -26,7 +26,7 @@ contract CusdcV3Wrapper is ICusdcV3Wrapper, WrappedERC20, CometHelpers {
     ICometRewards public immutable rewardsAddr;
     IERC20 public immutable rewardERC20;
 
-    mapping(address => uint64) public baseTrackingIndex; // uint64 to stay consistent with CometHelpers
+    mapping(address => uint64) public baseTrackingIndex; // uint64 for consistency with CometHelpers
     mapping(address => uint256) public baseTrackingAccrued; // uint256 to avoid overflow in L:199
     mapping(address => uint256) public rewardsClaimed;
 
--- a/contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
+++ b/contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
@@ -257,7 +257,8 @@ contract CusdcV3Wrapper is ICusdcV3Wrapper, WrappedERC20, CometHelpers {
 
         uint256 indexDelta = uint256(trackingSupplyIndex - baseTrackingIndex[account]);
         uint256 newBaseTrackingAccrued = baseTrackingAccrued[account] +
-            safe64((safe104(balanceOf(account)) * indexDelta) / TRACKING_INDEX_SCALE);
+            (safe104(balanceOf(account)) * indexDelta) /
+            TRACKING_INDEX_SCALE;
 
         uint256 claimed = rewardsClaimed[account];
         uint256 accrued = newBaseTrackingAccrued * RESCALE_FACTOR;
--- a/contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
+++ b/contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
@@ -26,8 +26,8 @@ contract CusdcV3Wrapper is ICusdcV3Wrapper, WrappedERC20, CometHelpers {
     ICometRewards public immutable rewardsAddr;
     IERC20 public immutable rewardERC20;
 
-    mapping(address => uint64) public baseTrackingIndex;
-    mapping(address => uint64) public baseTrackingAccrued;
+    mapping(address => uint64) public baseTrackingIndex; // keep uint64 to stay consistent with comet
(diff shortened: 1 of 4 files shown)


test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
```diff
@@ -598,6 +598,18 @@ describeFork('Wrapped CUSDCv3', () => {
       )
     })
 
+    it('regression test: able to claim rewards even when they are big without overflow', async () => {
+      // Nov 28 2023: uint64 math in CusdcV3Wrapper contract results in overflow when COMP rewards are even moderately large
+
+      const compToken = <ERC20Mock>await ethers.getContractAt('ERC20Mock', COMP)
+      expect(await compToken.balanceOf(wcusdcV3.address)).to.equal(0)
+      await advanceTime(1000)
+      await enableRewardsAccrual(cusdcV3, bn('2e18')) // enough to revert on uint64 implementation
```
pmckelvy1, line 607: i think we need to push it to 20e18, no?

jankjr: Yes

contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
```diff
@@ -26,8 +26,8 @@ contract CusdcV3Wrapper is ICusdcV3Wrapper, WrappedERC20, CometHelpers {
     ICometRewards public immutable rewardsAddr;
     IERC20 public immutable rewardERC20;
 
-    mapping(address => uint64) public baseTrackingIndex;
-    mapping(address => uint64) public baseTrackingAccrued;
+    mapping(address => uint64) public baseTrackingIndex; // uint64 to stay consistent with CometHelpers
```
julianmrodri, line 29: We need to fix the linting here by modifying the comment

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
julianmrodri: Confirmed the test fails if the change is reverted so seems to be good enough the example.

julianmrodri (commented):
The change LGTM

contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
```diff
@@ -286,9 +286,7 @@ contract CusdcV3Wrapper is ICusdcV3Wrapper, WrappedERC20, CometHelpers {
         (, uint64 trackingSupplyIndex) = getSupplyIndices();
         uint256 indexDelta = uint256(trackingSupplyIndex - baseTrackingIndex[account]);
 
-        baseTrackingAccrued[account] += safe64(
-            (safe104(accountBal) * indexDelta) / TRACKING_INDEX_SCALE
-        );
+        baseTrackingAccrued[account] += (safe104(accountBal) * indexDelta) / TRACKING_INDEX_SCALE;
```
julianmrodri, line 289: `houldnt we also change the `getRewardsOwed()` in L 259?

akshatmittal, line 289: Is there a reason to not just use full precision here?

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
akshatmittal, line 607: Can confirm this fails, so seems good. I also like the `2e18` example because it's under the `uint64` limit and hence highlights where the issue comes from.

akshatmittal (commented):
Added a few comments, looks good overall.

I think it's also worth changing the `RESCALE_FACTOR` to a `uint256`. (GitHub doesn't let me comment on the line itself)

test/plugins/individual-collateral/compoundv3/CusdcV3Wrapper.test.ts
tbrent: Yes I confirmed the test fails, so I don't see why it would need to be higher

pmckelvy1: yup nvm, this isn't the variable i thought it was.

contracts/plugins/assets/compoundv3/CusdcV3Wrapper.sol
tbrent: Agreed.

tbrent: What do you mean, full precision? The math here is in `uint256` and the product is before the division.

akshatmittal: Nvm, the comment is incorrect, I understood it wrong.

PR #1017, opened 2023-11-29
Invariant Monitor contract

julianmrodri:
* Monitor specific RTokens
* Meant to be used by a monitoring system
* Set as UUPS Proxy
* Backing Redeemable checks support: AAVE V2, AAVE V3, COMPOUND V2, COMPOUND V3., Stargate, FLUX, and Morpho AAVE V2.
* (need to pass the address of the ERC20 being used in the RToken as collateral - wrapped/static token)
*Includes deployment address and docs

Deployment task available:
* For the initial deployment: 
`npx hardhat deploy-facade-monitor --upgrade false --owner OWNER_ADDR --network mainnet` 

* For deploying just the new implementation to be used in upgrade:
`npx hardhat deploy-facade-monitor --upgrade true --network mainnet`


pmckelvy1:
im thinking its probably worth it to split this contract up into multiple.  it would be a lot easier to maintain each individual monitoring function if they need to be tweaked or upgraded (less gas costs to do individual deployments).
another thing to consider is using a proxy for each monitoring function contract, allowing them to updated individually.  this would have the benefit of not needing to change anything in hypernative if we need to do an update.
open to thoughts here as i'm not sure yet what the best path is.

pmckelvy1 (approved):

pmckelvy1:
> im thinking its probably worth it...

maybe for simplicity's sake, we just make this a proxy and use a single contract for now.  if we get to a point of wanting to make change or add functionality, we can reassess if a single contract is the best route, but for now, a single upgradeable proxy is probably easiest.

the only concern there is who is the admin? i think there is a msig we can use to approve upgrades.

julianmrodri:
@pmckelvy1  This is ready, included the morpho change, deployed and upgraded the monitors in Base and Mainnet

pmckelvy1 (approved):
lgtm

PR #1018, opened 2023-11-30
add scripts to upgrade tokens

jankjr:


pmckelvy1:
i think there is more noise in this pr than we need to include in the core repo, and its pretty much all one-off code.
I'll leave it open for another 24hr to see if anyone disagrees.

akshatmittal:
> i think there is more noise in this pr than we need to include in the core repo

I agree with @pmckelvy1, I'd prefer/recommend putting all of this stuff in a separate repo.

PR #1022, opened 2023-12-05
Add ability to reweight target basket

tbrent:
This turned out to be simpler than I thought because you don't actually need to forbid redemptions from previous basket nonces of different worths. I want the reviewer to think through this to confirm, though.

Only annoying part was we hit contract size limit and had to move to a slightly less readable way of applying the `governance` modifier. 

Regarding tests: `freshBasketHandler` was no longer required now that governance can make an RToken revaluable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,9 +10,13 @@ Upgrade BasketHandler
 
 ### Core Protocol Contracts
 
+New governance param added to `DeploymentParams`: `reweightable`
+
 - `BasketHandler` [+1 slot]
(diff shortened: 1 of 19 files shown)


tbrent:
> This PR is much smaller than I expected, haha.

It was much larger before

tbrent:
Holding off on reviewing/merging until we decide on immutable/mutable

tbrent:
@akshatmittal ready for review

akshatmittal (changes requested):

tbrent:
How would that work? Has to be set in the `init()` not constructor.

akshatmittal (approved):

PR #1024, opened 2023-12-07
cleanup slithering, add pipeline job.

pmckelvy1:


pmckelvy1:
doing a full cleanup and validation of all slither results is a much larger ticket, maybe 13 at least.

this pr adds a job to the pipeline and cleans up some the current HIGH issues raised by slither (verified to be non-issues), as well as a few mediums. it also excludes LOW, INFORMATIONAL, and OBSERVATIONS form the output.

contracts/facade/FacadeAct.sol
```diff
@@ -286,3 +287,4 @@ contract FacadeAct is IFacadeAct, Multicall {
         revert("unrecognized version");
     }
 }
+// slither-disable-end divide-before-multiply
```
tbrent, line 290: I think the `divide-before-multiply` was unintentional

tbrent (approved):
One small nit but other than that LGTM. It's unfortunate there's no detector for implicit up-casts, or something similar. Still, this improves CI.

tbrent (dismissed):

tbrent (approved):
LGTM

PR #1025, opened 2023-12-09
Gas-optimization: remove unnecessary main.* calls

tbrent:
There were two `main.*` usages in P1 we didn't mean to leave in there.


tbrent:
This should go in after #1022

akshatmittal (approved):

PR #1026, opened 2023-12-12
fix sfrxETH Collateral, add new oracle

pmckelvy1:


contracts/plugins/assets/frax-eth/SFraxEthCollateral.sol
```diff
@@ -49,22 +60,21 @@ contract SFraxEthCollateral is AppreciatingFiatCollateral {
             uint192 pegPrice
         )
     {
-        // {UoA/tok} = {UoA/target} * {ref/tok} * {target/ref} (1)
-        uint192 p = chainlinkFeed.price(oracleTimeout).mul(_underlyingRefPerTok());
+        uint192 targetPerTok = targetPerTokChainlinkFeed.price(targetPerTokChainlinkTimeout);
+        // {UoA/tok} = {UoA/target} * {target/tok}
```
julianmrodri, line 64: Would leave an empty line before this line as with other plugins

contracts/plugins/assets/frax-eth/SFraxEthCollateral.sol
```diff
@@ -23,17 +24,27 @@ import "./vendor/IsfrxEth.sol";
  */
 contract SFraxEthCollateral is AppreciatingFiatCollateral {
     using OracleLib for AggregatorV3Interface;
+    using FraxOracleLib for FraxAggregatorV3Interface;
```
julianmrodri, line 27: Should we remove the comment that "THIS PLUGIN IS NOT READY TO BE USED IN PRODUCTION"?

contracts/plugins/assets/frax-eth/SFraxEthCollateral.sol
```diff
@@ -23,17 +24,27 @@ import "./vendor/IsfrxEth.sol";
...
-    // solhint-disable no-empty-blocks
-    /// @param config.chainlinkFeed Feed units: {UoA/target}
-    constructor(CollateralConfig memory config, uint192 revenueHiding)
-        AppreciatingFiatCollateral(config, revenueHiding)
-    {
+    FraxAggregatorV3Interface public immutable targetPerTokChainlinkFeed; // {target/tok}
+    uint48 public immutable targetPerTokChainlinkTimeout;
+
+    /// @param config.chainlinkFeed {UoA/target} price of ETH in USD terms
+    /// @param _targetPerTokChainlinkFeed {target/tok} price of cbETH in ETH terms
```
julianmrodri, line 34: `cbEth` should be replaced right?

contracts/plugins/mocks/ChainlinkMock.sol
```diff
@@ -49,6 +51,17 @@ contract MockV3Aggregator is AggregatorV3Interface {
         latestAnsweredRound = latestRound;
     }
 
+    // used by Frax oracl
```
julianmrodri, line 54: typo `oracle`

test/plugins/individual-collateral/frax-eth/SFrxEthTestSuite.test.ts
```diff
@@ -82,13 +93,18 @@ export const deployCollateral = async (opts: CollateralOpts = {}): Promise<TestI
...
+    opts.targetPerTokChainlinkFeed ?? SFRXETH_ETH_PRICE_FEED,
+    opts.targetPerTokChainlinkTimeout ?? ORACLE_TIMEOUT,
     { gasLimit: 2000000000 }
   )
   await collateral.deployed()
 
   // Push forward chainlink feed
   await pushOracleForward(opts.chainlinkFeed!)
-
+  // await pushOracleForward(opts.targetPerTokChainlinkFeed!)
```
julianmrodri, line 104: can remove this line probably

julianmrodri (approved):
LGTM overall. We should update the `README.md` for this plugin to reflect that there is an oracle now (final section). Also at the top we should probably remove that this is not ready for PROD in both the contract and `README`.

pmckelvy1:
realized it made more sense to use the frxETH/ETH oracle and not sfrxETH/ETH.

also now realizing that FRAX does not give a % diff SLA in their oracle updates, so if we want to use this, we might need to hookup a keeper.

tbrent:
Introducing the new `itChecksForTargetPerRefDefaultUp` seems necessary. Didn't like it at first but don't see a better way to do it.

Not sure about the oracleError for the EMA. Here we are implicitly saying it is 0. The case I'm considering is what should `price()` be for a collateral _entirely_ priced by its EMA. I think 0 would be an issue? Is there some type of standard deviation we can use, or something else we can derive from the windowing / exponent in the EMA?

tbrent:
Moving to 3.2.0. Was not part of mitigation review with Trust.

tbrent:
Deployment scripts to follow in a second PR. The expectation is `oracleError` will include the curve pool trading fee.

PR #1027, opened 2023-12-14
Feature/dutch auction cb bid

jankjr:


test/utils/bidOnTrade.ts
```diff
@@ -0,0 +1,36 @@
+import { Signer, constants } from 'ethers'
```
jankjr, line 1: I'm not a fan of this module, I'll rewrite tomorrow before anyone wastes time reviewing it.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+pragma solidity 0.8.19;
+
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IDutchTradeCallee, TradeStatus, DutchTrade } from "../trading/DutchTrade.sol";
+import { IMain } from "../../interfaces/IMain.sol";
+
+contract DutchTradeRouter is IDutchTradeCallee {
+    using SafeERC20 for IERC20;
+    struct Bid {
```
jankjr, line 11: Will update this struct as it bothers me I chose not to follow the same naming scheme as the DutchTrade

tbrent (commented):
Everywhere in the `DutchTradeRouter` let's use our unit notation to document the type of each variable: https://github.com/reserve-protocol/protocol/blob/master/docs/solidity-style.md#units-in-comments

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+    function _sendBalanceTo(IERC20 token, address to) internal {
+        uint256 bal = token.balanceOf(address(this));
+        if (bal == 0) {
+            return;
+        }
+        token.safeTransfer(to, bal);
+    }
+
+    // Places a bid on a Dutch auction
+    // Method will dynamically pull funds from msg.sender if needed
```
tbrent, line 87: I don't think we want to use "if needed" logic here given that the router will be used by many different parties. It should always transfer in exactly the number of tokens needed, imo, else we're just asking for a low-information party to send tokens to the contract before getting frontrun by an MEV bot during the actual bid.

jankjr: Yeah I'll remove that branch. It was a leftover from when I envisioned supporting placing multiple bids at the same time via the UI.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+    event BidPlaced(
+        IMain main,
+        DutchTrade trade,
+        address bidder,
+        IERC20 sellToken,
+        uint256 soldAmt,
+        IERC20 buyToken,
+        uint256 boughtAmt
+    );
+    DutchTrade private _currentTrade = DutchTrade(address(0));
```
akshatmittal, line 41: Unnecessary initialization.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+
+    /// Place a bid on an OPEN dutch auction
+    /// @param trade The DutchTrade to bid on
+    /// @param recipient The recipient of the tokens out
+    /// @dev Requires msg.sender has sufficient approval on the tokenIn with router
+    /// @dev Requires msg.sender has sufficient balance on the tokenIn
+    function bid(DutchTrade trade, address recipient) external returns (Bid memory) {
+        Bid memory out = Bid({
```
akshatmittal, line 49: Unnecessary initialization.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+        require(_currentTrade == DutchTrade(address(0)), "already bidding");
+        require(trade.status() == TradeStatus.OPEN, "trade not open");
+        out.trade = trade;
+        out.buyToken = IERC20(trade.buy());
+        out.sellToken = IERC20(trade.sell());
+        out.buyAmt = trade.bidAmount(block.number);
+
+        uint256 currentBalance = out.buyToken.balanceOf(address(this));
+        if (currentBalance < out.buyAmt) {
+            out.buyToken.safeTransferFrom(bidder, address(this), out.buyAmt - currentBalance);
```
akshatmittal, line 104: Feels weird to me, it's probably best to pull the exact number of tokens.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,125 @@
...
+        }
+        uint256 sellAmt = out.sellToken.balanceOf(address(this));
+        _currentTrade = trade;
```
akshatmittal, line 107: Let's set the value of `_currentTrade` a little higher in the chain to avoid ERC777 style issues.

akshatmittal (approved):
Looks good to me! Recommended some small changes below, logic looks sound!

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -183,12 +192,16 @@ contract DutchTrade is ITrade {
...
+    /// Bid with callback for the auction lot at the current price;
+    ///  Sold funds are sent back to the callee, callee.dutchTradeCallback(...) is invoked
+    ///  balance of buy token must increase by bidAmount(current block) after callback
+    ///  Trade is settled atomically via a callback
+    ///
+    /// @param data {bytes} The data to pass to the callback
+    /// @dev Caller must implement IDutchTradeCallee
     /// @return amountIn {qBuyTok} The quantity of tokens the bidder paid
-    function bid() external returns (uint256 amountIn) {
+    function bid(bytes calldata data) external returns (uint256 amountIn) {
```
pmckelvy1, line 203: we should probably add nonreentrant to this?

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -206,6 +218,13 @@ contract DutchTrade is ITrade {
         if (price > bestPrice.mul(ONE_POINT_FIVE, CEIL)) {
             broker.reportViolation();
         }
+        sell.safeTransfer(bidder, lot()); // {qSellTok}
+        uint256 balanceBefore = buy.balanceOf(address(this));
+        IDutchTradeCallee(bidder).dutchTradeCallback(msg.sender, address(buy), amountIn, data);
```
pmckelvy1, line 223: use `bidder` instead of `msg.sender`

pmckelvy1 (changes requested):

contracts/plugins/trading/DutchTrade.sol
akshatmittal: I thought of this as well, but I think this function is already non-reentrant because of the `bidder == address(0)` check at the very beginning.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,110 @@
...
+        out.sellToken = IERC20(trade.sell());
+        out.buyAmt = trade.bidAmount(block.number);
+        out.buyToken.safeTransferFrom(bidder, address(this), out.buyAmt);
+
+        uint256 sellAmt = out.sellToken.balanceOf(address(this));
+        uint256 expectedSellAmt = trade.lot();
+        trade.bid(new bytes(0));
+
+        sellAmt = out.sellToken.balanceOf(address(this)) - sellAmt;
+        require(sellAmt >= expectedSellAmt, "insufficient amount out");
```
pmckelvy1, line 96: comment units

pmckelvy1 (changes requested):

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -197,7 +211,6 @@ contract DutchTrade is ITrade {
 
         // Transfer in buy tokens
```
tbrent, line 212: This comment should change, probably "Mark bid received" or something

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,111 @@
...
+    /// @param buyToken The token DutchTrade is expecting to receive
+    /// @param buyAmount The amt the DutchTrade is expecting to receive {qBuyToken}
+    /// @notice Data is not used here
+    function dutchTradeCallback(
+        address caller,
+        address buyToken,
+        uint256 buyAmount,
+        bytes calldata
+    ) external {
+        require(caller == address(this), "Invalid caller");
```
tbrent, line 66: I'm not sure I see the point of the `caller` variable. Whichever contract is calling this router contract (DutchTrade or not) can pass in whatever value they want. I think the `msg.sender` check is the only one we need.

contracts/plugins/trading/DutchTradeRouter.sol
```diff
@@ -0,0 +1,111 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+pragma solidity 0.8.19;
+
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { IDutchTradeCallee, TradeStatus, DutchTrade } from "../trading/DutchTrade.sol";
+import { IMain } from "../../interfaces/IMain.sol";
+
+contract DutchTradeRouter is IDutchTradeCallee {
```
tbrent, line 9: Let's document the class and give a brief description of how to use the contract as a bidder here

tbrent (commented):
Nice tests!

Left a comment about a variable I'm not sure is needed in the callback function. 

Let's also touch `docs/mev.md` and at least point them towards the existence of this router, though I think that document should remain describing how to bid against a `DutchTrade` directly.

pmckelvy1:
1/9/24:
- add 2nd codepath instead of change bid() interface
- bid() should be the 3.1.0 version
- bidWithCallback() is the new function added in this pr

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -200,7 +207,7 @@ contract DutchTrade is ITrade {
     /// @param data {bytes} The data to pass to the callback
     /// @dev Caller must implement IDutchTradeCallee
     /// @return amountIn {qBuyTok} The quantity of tokens the bidder paid
-    function bid(bytes calldata data) external returns (uint256 amountIn) {
+    function bidCb(bytes calldata data) external returns (uint256 amountIn) {
```
akshatmittal, line 210: Can we call this `bitWithCallback` instead? Feels better to me.

akshatmittal (changes requested):

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -73,6 +88,7 @@ contract DutchTrade is ITrade {
     using SafeERC20 for IERC20Metadata;
 
     TradeKind public constant KIND = TradeKind.DUTCH_AUCTION;
+    BidType public bidType = BidType.NONE;
```
tbrent, line 91: This line is slightly weird because the contract is used as a clone. It's fine, because `bidType` will be 0 in the proxies on deployment, but having the assignment at all feels misleading. I'd rather leave the assignment out and throw a comment in there `// = BidType.NONE`. Bit of a nit I know

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -183,6 +199,50 @@ contract DutchTrade is ITrade {
...
+    ///  Sold funds are sent back to the callee, callee.dutchTradeCallback(...) is invoked
+    ///  balance of buy token must increase by bidAmount(current block) after callback
+    ///  Trade is settled atomically via a callback
+    ///
+    /// @param data {bytes} The data to pass to the callback
+    /// @dev Caller must implement IDutchTradeCallee
+    /// @return amountIn {qBuyTok} The quantity of tokens the bidder paid
+    function bidWithCallback(bytes calldata data) external returns (uint256 amountIn) {
+        require(bidder == address(0), "bid already received");
+        // {buyTok/sellTok}
```
tbrent, line 212: let's put a line between 211 and 212 if we're going to do it in `bid()`. Everything should be parallel as much as possible.

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -183,6 +199,50 @@ contract DutchTrade is ITrade {
...
+        uint192 price = _price(block.number); // enforces auction ongoing
+
+        // {qBuyTok}
+        amountIn = _bidAmount(price);
+
+        // Mark bidder
+        bidder = msg.sender;
```
tbrent, line 219: this section is not parallel between the two functions. I think all the assignments should be grouped as one section and then the `safeTransferFrom()` on line 261 should stand on its own.

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -223,13 +284,15 @@ contract DutchTrade is ITrade {
         returns (uint256 soldAmt, uint256 boughtAmt)
     {
         require(msg.sender == address(origin), "only origin can settle");
+        if (bidder == address(0)) {
```
tbrent, line 287: these 3 lines can be simplified to `require(bidder != address(0) || block.number > endBlock, "auction not over");`

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -183,6 +199,50 @@ contract DutchTrade is ITrade {
...
+        );
+
+        // settle() via callback
+        origin.settleTrade(sell);
+
+        // confirm callback succeeded
+        assert(status == TradeStatus.CLOSED);
+    }
+
     /// Bid for the auction lot at the current price; settling atomically via a callback
```
tbrent, line 246: This is now confusing, because the callback it is referring to is the `origin.settleTrade()` call. We should distinguish between the bidder callback and protocol callback in this contract.

tbrent (commented):
The `bidType` state tracking is uglier than I'd like it to be, but I don't see a good alternative. I tried a few things and none of them are better.

I did leave some fussy comments, though. I think it's important for this contract to be as clear as possible, and there were a few things around the edges I thought detracted from that.

Also, I think we should swap the implementation order of `bid()` and `bidWithCallback()`, I think it makes more sense for `bid()` to come first.

contracts/plugins/trading/DutchTrade.sol
jankjr: Reordered the implementations to correctly reflect each other.

pmckelvy1 (approved):
lgtm

tbrent (approved):
LGTM

PR #1029, opened 2023-12-20
Prevent Stargate wrapper from reverting on failed claimRewards

jankjr:
Wrap the .deposit call in a try {} catch {}


pmckelvy1 (changes requested):
probably worth adding a similar try/catch in _beforeWithdraw. we do a similar check there, but we obv can't rely on allocPoint or totalAllocPoint

tbrent:
LGTM

PR #1030, opened 2023-12-21
remove unused function on IMorpho interface

tbrent:


pmckelvy1:
should we base this against 3.1.0 since there are other changes in that branch as well?

PR #1032, opened 2024-01-11
TRUST-M3: Fix Storage Gap

julianmrodri:
* Fixes storage gap in BackingManager P1, which was decreased by 2 instead of 1.
* Validated the change with our own `validate-upgrade` task to confirm this was OK.

image


pmckelvy1 (approved):
lgtm

PR #1033, opened 2024-01-11
TRUST-L8: Remove unnecessary cast

julianmrodri:
Removes an unnecessary cast. Double checked these were the only two lines that required the fix.


pmckelvy1 (approved):
lgtm

PR #1040, opened 2024-01-17
Bid tests - complete tests for both bidding types

julianmrodri:
* Adds bid types to the Broker extreme tests
* Prepares 3 tests in Revenues to run for both Bid Types


tbrent:
Nice catch, LGTM

PR #1042, opened 2024-01-18
UnstakingStarted event fix

tbrent:
Fix use of `era` in emitted `UnstakingStarted`; use `draftEra`

The changes to the P1 core contracts are so obviously safe I think this can go in 3.1.0. 

Also changes `FacadeRead.pendingUnstakings()` to take the draftEra.

### DIFF
--- a/contracts/p1/StRSR.sol
+++ b/contracts/p1/StRSR.sol
@@ -76,14 +76,14 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
     // === Financial State: Drafts ===
     // Era. If drafts get wiped out due to RSR seizure, increment the era to zero draft values.
     // Only ever directly written by beginDraftEra()
-    uint256 internal draftEra;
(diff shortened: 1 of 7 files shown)


contracts/p1/StRSR.sol
```diff
@@ -76,15 +76,15 @@ abstract contract StRSRP1 is Initializable, ComponentP1, IStRSR, EIP712Upgradeab
     // === Financial State: Drafts ===
     // Era. If drafts get wiped out due to RSR seizure, increment the era to zero draft values.
     // Only ever directly written by beginDraftEra()
-    uint256 internal draftEra;
+    uint256 internal draftEra; // {draftEra}
```
akshatmittal, line 79: Might as well make this public?

akshatmittal (approved):

contracts/p1/StRSR.sol
tbrent: wasn't quite ready for your review, you may want to make a full pass again

akshatmittal (approved):

PR #1049, opened 2024-01-30
TRST-M-1b: Add basket reweighting for index tokens

tbrent:
This PR adds automatic USD re-weighting for index RTokens (`reweightable = true`) to `setPrimeBasket()`. An additional `forceSetPrimeBasket()` is added to bypass the re-weighting, since re-weighting comes with the additional assumption that the basket is currently SOUND. 

There are tests in this PR as well as fuzz tests for `BasketLib.normalizeByPrice()` over in the `protocol-fuzzing` repo
https://github.com/reserve-protocol/protocol-fuzzing/blob/main/test/protocol/BasketLib.t.sol

Minor changes: also streamlined some naming in `requireConstantConfigTargets`. The `_` prefix for `_targetAmts` is actually a great way to indicate "this variable is different than the others it's more like scratch storage". (btw the diff is misleading in `requireConstantConfigTargets`)


tbrent:
I've added an alternative flow: `forceSetPrimeBasket()` that governance can use in times of crisis to skip reweighting the basket unit to be a constant USD. Since normalizing target amounts requires a SOUND previous basket, governance may sometimes want to forcibly set the prime basket, but the default behavior should be restrictive.

tbrent:
@pmckelvy1 @akshatmittal I'd like both your eyes on this one

contracts/p1/BasketHandler.sol
```diff
@@ -197,21 +214,39 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
     //   config'.erc20s = erc20s
     //   config'.targetAmts[erc20s[i]] = targetAmts[i], for i from 0 to erc20s.length-1
     //   config'.targetNames[e] = assetRegistry.toColl(e).targetName, for e in erc20s
-    function setPrimeBasket(IERC20[] calldata erc20s, uint192[] calldata targetAmts) external {
+    function _setPrimeBasket(
+        IERC20[] calldata erc20s,
+        uint192[] memory targetAmts,
```
pmckelvy1, line 219: calldata?

test/Main.test.ts
```diff
@@ -1752,77 +1764,122 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
...
       )
     })
 
     it('Should not allow to set prime Basket with a zero amount', async () => {
+      await expect(indexBH.connect(owner).setPrimeBasket([token0.address], [0])).to.be.revertedWith(
+        'invalid target amount; must be nonzero'
+      )
+      await expect(
+        basketHandler.connect(owner).setPrimeBasket([token0.address], [0])
+      ).to.be.revertedWith('missing target weights')
```
pmckelvy1, line 1877: this revert message is different which makes me think maybe theres a different way we want to test this for non-reweightable baskets.
set token0 to the same amt but add token1 (or whatever token has same target) and set it to 0?

pmckelvy1: also not entirely sure it matters, this test may be sufficient. it just smells weird...

test/Main.test.ts
```diff
@@ -1833,6 +1890,12 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
           [token0.address, token1.address, token2.address, token3.address],
           [fp('0.25'), fp('0.25'), fp('0.25'), fp('0.25')]
         )
+      await basketHandler
```
pmckelvy1, line 1893: do we need to test this for indexBH as well?

test/Main.test.ts
```diff
@@ -1855,6 +1918,14 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
       await expect(basketHandler.connect(owner).setPrimeBasket([token2.address], [fp('1')]))
         .to.emit(basketHandler, 'PrimeBasketSet')
         .withArgs([token2.address], [fp('1')], [ethers.utils.formatBytes32String('USD')])
+
+      await expect(basketHandler.connect(owner).forceSetPrimeBasket([token1.address], [fp('1')]))
```
pmckelvy1, line 1922: should we test these for indexBH as well?

test/Main.test.ts
```diff
@@ -1926,13 +2022,24 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
...
+          .forceSetPrimeBasket([token0.address, rToken.address], [fp('0.5'), fp('0.5')])
+      ).to.be.revertedWith('invalid collateral')
+      await expect(
+        basketHandler
+          .connect(owner)
+          .forceSetPrimeBasket([token0.address, rToken.address], [fp('0.5'), fp('0.5')])
+      ).to.be.revertedWith('invalid collateral')
     })
 
     it('Should revert if target has been changed in asset registry', async () => {
```
pmckelvy1, line 2045: should this even be allowed?

pmckelvy1 (changes requested):
this pr is the definition of "hold on, there is probably a simpler way to do this..."
🔥

contracts/p1/BasketHandler.sol
tbrent: can't be -- it gets mutated down below, conditionally

test/Main.test.ts
tbrent: yes...

test/Main.test.ts
tbrent: hard to prevent because can break `swapRegistered()` into `unregister()` and `register()`

pmckelvy1 (approved):
lgtm

akshatmittal (approved):

PR #1050, opened 2024-01-30
TRST-H-1: YearnV2CurveFiatCollateral

akshatmittal:
Fixes:
* TRST-H-1

Confirmed the test fails for the previous implementation.

### DIFF
--- a/contracts/plugins/assets/yearnv2/YearnV2CurveFiatCollateral.sol
+++ b/contracts/plugins/assets/yearnv2/YearnV2CurveFiatCollateral.sol
@@ -4,10 +4,11 @@ pragma solidity 0.8.19;
 import "../curve/CurveStableCollateral.sol";
 
 interface IPricePerShareHelper {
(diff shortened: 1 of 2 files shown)


contracts/plugins/assets/yearnv2/YearnV2CurveFiatCollateral.sol
```diff
@@ -100,12 +101,12 @@ contract YearnV2CurveFiatCollateral is CurveStableCollateral {
     /// @return {LP token/tok}
     function _pricePerShare() internal view returns (uint192) {
```
tbrent, line 102: Thoughts on what Patrick raised, instead? 

```
return _safeWrap(pricePerShareHelper.sharesToAmount(address(erc20), 10**lpToken.decimals()));
```

tbrent (approved):
I'm fine with this as-is (congrats on testing via storage manipulation btw) but wanted to drop the possibly simpler alternative

PR #1051, opened 2024-01-31
TRST-QA-3: Gap Sizes

akshatmittal:


akshatmittal:
@tbrent I'd like you to confirm these gap sizes, I also found the gap size on StRSRP1 to be incorrect in addition to the other contracts.

tbrent:
Interesting. Will look into. Wrong gap size on `StRSRP1` would actually matter because of `StRSRVotes`.

contracts/p1/RToken.sol
```diff
@@ -530,5 +530,5 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
      * variables without shifting down storage in the inheritance chain.
      * See https://docs.openzeppelin.com/contracts/4.x/upgradeable#storage_gaps
      */
-    uint256[42] private __gap;
+    uint256[36] private __gap;
```
tbrent, line 533: Man...why EIP712Upgradeable using 52 slots :(

tbrent:
Trust confirmed `BasketHandler` is 28

tbrent (approved):

PR #1057, opened 2024-02-06
Make LP token pricing less manipulable

tbrent:
Previous pool pricing method was overly reliant on spot balances, leading to the ability to manipulate `price()` upwards for LP token collateral simply by imbalancing the pool in a sandwich attack. This PR removes any usage of the pool balances from the pricing algorithm, preferring instead to completely rely on oracles. We might consider merging the oracle and balance data if we had TWAPs available, but the Curve pools we are interested in do not. With only spot balances available, the only non-manipulable pricing methods left are ones that do not consider spot balances at all.

Aside: `FixLib.powu()` tests were not actually running, and we were testing cases that should revert

### DIFF
--- a/contracts/libraries/Fixed.sol
+++ b/contracts/libraries/Fixed.sol
@@ -329,10 +329,6 @@ library FixLib {
         return _safeWrap(result / FIX_SCALE);
     }
 
-    function sqrt(uint192 x) internal pure returns (uint192) {
-        return _safeWrap(sqrt256(x * FIX_ONE_256));
-    }
-
     /// Comparison operators...
     function lt(uint192 x, uint192 y) internal pure returns (bool) {
         return x < y;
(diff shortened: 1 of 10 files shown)


tbrent:
@akshatmittal when we do USDC+/USDC, we can grab the `sqrt()` function I implemented here at first and then deleted. We'll need it for the `underlyingRefPerTok()` which should come out to a product of the virtual price with the square root of the USDC+/USDC exchange rate.

akshatmittal:
Given the changes here, might be worth simplifying the `PoolTokens` class as well? We're no longer using a large portion of it.

Additionally the example used in the `refPrice` implementation is incorrect, how much the pool moves is a factor of it's `A` and multiplier (For example the [mkUSD/USDC](https://curve.fi/#/ethereum/pools/factory-stable-ng-17/deposit) pools is currently 74:26), that being said, the implementation to find the _minimum_ value of the pool looks correct. We should probably add additional documentation around that though, this will always price them to the lowest so the actual redemption/issuance can be at higher than expected.

I'd ideally also want to run some simulations on this.

tbrent:
I think I'm coming around to preferring our previous LP token pricing method. It's manipulable but not very much so, and I'm not sure we can do better. To discuss, but might just closeout this PR

tbrent:
Closing in favor of the comment in commit c02cb71

PR #1070, opened 2024-02-13
use maximum oracle timeout for the price decay delay

tbrent:
The `price()` function in `Asset.sol` assumes it has the maximum oracle timeout value but it does not. We have at least 1 collateral that has the main `config.oracleTimeout` as its min and one of the other timeouts as its max. So I've implemented an internal function `_decayDelay()` that each plugin should override if it is using more than `config.oracleTimeout`.


tbrent:
@akshatmittal to experiment with pulling the timeout out of the `StalePrice()` error in `OracleLib`

tbrent:
Collapsing into 1 PR, will use that CI

PR #1072, opened 2024-02-14
add facade function to query issuable RToken using amounts

tbrent:


contracts/facade/FacadeRead.sol
```diff
@@ -28,26 +28,46 @@ contract FacadeRead is IFacadeRead {
...
-        BasketRange memory basketsHeld = main.basketHandler().basketsHeldBy(account);
-        uint192 needed = rToken.basketsNeeded();
-
-        int8 decimals = int8(rToken.decimals());
-
-        // return {qRTok} = {BU} * {(1 RToken) qRTok/BU)}
-        if (needed.eq(FIX_ZERO)) return basketsHeld.bottom.shiftl_toUint(decimals);
+        // Get basket ERC20s
+        IBasketHandler bh = main.basketHandler();
+        if (!bh.isReady()) return 0;
```
akshatmittal, line 55: Do we need this if we're just quoting here?

contracts/facade/FacadeRead.sol
```diff
@@ -28,26 +28,46 @@ contract FacadeRead is IFacadeRead {
...
+        (address[] memory erc20s, uint256[] memory quantities) = bh.quote(FIX_ONE, CEIL);
 
-        uint192 totalSupply = shiftl_toFix(rToken.totalSupply(), -decimals); // {rTok}
+        // Compute how many baskets we can mint with the collateral amounts
+        uint192 baskets = type(uint192).max;
+        for (uint256 i = 0; i < erc20s.length; ++i) {
+            // {BU} = {tok} / {tok/BU}
+            uint192 inBUs = divuu(amounts[i], quantities[i]); // FLOOR
```
akshatmittal, line 62: Given this, should add a comment that the quantities must be in the same order as the erc20s in the basket.

akshatmittal (changes requested):

contracts/facade/FacadeRead.sol
tbrent: No we don't...

contracts/facade/FacadeRead.sol
tbrent: Good point

akshatmittal (approved):

PR #1073, opened 2024-02-14
Singleton Facade

tbrent:


contracts/facade/facets/ActFacet.sol
```diff
@@ -4,24 +4,25 @@ pragma solidity 0.8.19;
...
- * @notice A Facade to help batch compound actions that cannot be done from an EOA, solely.
+ * @title ActFacet
+ * @notice
+ *   Facet to help batch compound actions that cannot be done from an EOA, solely.
  *   Compatible with both 2.1.0 and ^3.0.0 RTokens.
+ * @custom:static-call - Use ethers callStatic() to get result after update; do not execute
  */
 // slither-disable-start
-contract FacadeAct is IFacadeAct, Multicall {
+contract ActFacet is IActFacet, Multicall {
```
akshatmittal, line 20: Is Facet the right thing to call them? I feels like it might confuse people that it's a Diamond.

tbrent: I don't find it problematic for people to have that confusion. If you're consuming the Facade then the Diamond analogy has the correct security model, and you don't really care about the details anyway. Open to proposals though, it's definitely true that the `Facade`/`Facet` naming is a product of history and there might be better options out there.

tbrent: `Face` is an interesting alternative to `Facet` that has practically the same meaning but is distinctly non-Diamond

tbrent: The other option is each individual implementation contract becomes a `FacadeX` and the proxy contract is renamed.

tbrent:
Naming approved

PR #1074, opened 2024-02-16
Revert "TRST-QA-3: Gap Sizes"

pmckelvy1:
Reverts reserve-protocol/protocol#1051

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ scripts/addresses/31337*
 
 *.orig
 .idea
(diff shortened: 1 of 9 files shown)


julianmrodri:
All validations pass on the OZ plugin

image

julianmrodri (approved):
LGTM fixes the validations so effectively rollsback the change

tbrent:
I think we should add a comment to all the non-StRSR contracts providing context

PR #1084, opened 2024-03-03
Arbitrum

tbrent:
- NetworkConfigLib changes required because `block.number` gives the L1 blocknumber
- StRSRP1 bumped over contract limit: error string changes
- DutchTrade altered to use time instead of blocks


contracts/libraries/NetworkConfigLib.sol
```diff
@@ -19,8 +26,23 @@ library NetworkConfigLib {
             return 12; // Ethereum PoS, Goerli, HH (tests)
         } else if (chainId == 8453 || chainId == 84531) {
             return 2; // Base, Base Goerli
+        } else if (chainId == 42161 || chainId == 421614) {
+            return 1; // round up to 1 even though Arbitrum is ~0.26s
```
akshatmittal, line 30: Does this just break how we use it though?

PR #1086, opened 2024-03-06
Version trading plugins

tbrent:
- Add `version() returns (string memory)` getter to `DutchTrade` and ` GnosisTrade`
- Set version to "3.3.0"


tbrent:
@akshatmittal to confirm: we should keep the `VersionedAsset` mixin but NOT create a new `VersionedTradingPlugin` or something similar

akshatmittal (approved):

akshatmittal:
@tbrent Seems ok, I'm good with that.

PR #1088, opened 2024-03-07
TRST-M-1 mit review: trackCollateralization() inside trackStatus()

tbrent:


akshatmittal (approved):
Also confirmed this touches the redemption paths.

PR #1098, opened 2024-03-20
Meta morpho scripts

tbrent:


scripts/deployment/phase2-assets/collaterals/deploy_bbusdt.ts
```diff
@@ -0,0 +1,91 @@
...
+
+  const collateral = <MetaMorphoFiatCollateral>await MetaMorphoFiatCollateralFactory.connect(
+    deployer
+  ).deploy(
+    {
+      priceTimeout: PRICE_TIMEOUT.toString(),
+      chainlinkFeed: USDT_USD_FEED,
+      oracleError: USDT_ORACLE_ERROR.toString(),
+      erc20: networkConfig[chainId].tokens.bbUSDT,
+      maxTradeVolume: fp('1e6').toString(), // $7.5m vault
```
akshatmittal, line 63: I think we should be using _much_ smaller numbers for this. The vaults are currently too small to sustain these.

akshatmittal (changes requested):

scripts/deployment/phase2-assets/collaterals/deploy_bbusdt.ts
tbrent: Ok I think $500k can make sense, but much lower and things become unusable on mainnet

tbrent: And <$500k for pyUSD, since it has $1.7m vault

akshatmittal: That sounds good to me. Probably $250k for smaller vaults and $500k for every other.

akshatmittal (approved):

PR #1103, opened 2024-03-28
Make StaticATokenV3LM.claimRewards() work for markets without rewards

tbrent:
Found an issue that could result in `claimRewards()` reverting for markets without rewards. 

TODO regression test


tbrent:
I wanted to open this as its own PR first to see if this breaks anything

PR #1109, opened 2024-04-02
Patch facade maxIssuable + maxIssuableByAmounts

tbrent:


tbrent:
I swear in my IDE the file showed `MaxIssuableAmounts.sol` but on the command line and here it was `maxIssuableAmounts.sol`. Drove me crazy for half an hour...

contracts/interfaces/IMaxIssuableFacet.sol
```diff
@@ -0,0 +1,25 @@
...
+
+import "./IRToken.sol";
+
+/**
+ * @title IMaxIssuableFacet
+ * @notice Facet for maxIssuable + maxIssuableByAmounts
+ *
+ * - @custom:static-call - Use ethers callStatic() in order to get result after update
+v */
+interface IMaxIssuableFacet {
```
akshatmittal, line 12: Don't create interfaces for it.

contracts/facade/facets/ReadFacet.sol
```diff
@@ -21,56 +22,11 @@ import "../../p1/StRSRVotes.sol";
  * @custom:static-call - Use ethers callStatic() to get result after update; do not execute
  */
 // slither-disable-start
-contract ReadFacet is IReadFacet {
+contract ReadFacet is MaxIssuableFacet, IReadFacet {
```
akshatmittal, line 25: This is not necessary, each facet can be it's own thing.

tbrent: it's nice for testing at the moment. it means we can just test the ReadFacet and not have to track a bunch of individual facets

PR #1111, opened 2024-04-05
StakeDAO recursive collateral

tbrent:
Context: Intermediate `sqrt()` work got merged into the base branch already

Still have 2 custom tests I want to write


contracts/libraries/Fixed.sol
```diff
@@ -675,4 +679,94 @@ function fullMul(uint256 x, uint256 y) pure returns (uint256 hi, uint256 lo) {
...
+///
+/// @dev See https://en.wikipedia.org/wiki/Methods_of_computing_square_roots#Babylonian_method.
+///
+/// Notes:
+/// - If x is not a perfect square, the result is rounded down.
+/// - Credits to OpenZeppelin for the explanations in comments below.
+///
+/// @param x The uint256 number for which to calculate the square root.
+/// @return result The result as a uint256.
+function sqrt256(uint256 x) pure returns (uint256 result) {
```
akshatmittal, line 694: I'm assuming this function is correct, not verifying the implementation.

contracts/plugins/assets/curve/CurveRecursiveCollateral.sol
```diff
@@ -0,0 +1,192 @@
...
+        // {ref/tok} = {ref/lpToken} = {lpToken@t=0/lpToken} * {1} * 2{ref/lpToken@t=0}
+        return virtualPrice.mul(rTokenRate.sqrt()).mulu(2); // LP token worth twice as much
+    }
+
+    // === Internal ===
+
+    // Override this later to implement non-standard recursive pools
+    function _anyDepeggedInPool() internal view virtual override returns (bool) {
+        // Check reference token
+        try this.tokenPrice(0) returns (uint192 low, uint192 high) {
```
akshatmittal, line 173: Probably needs a comment saying that this only works where the first token is the paired token and the second token is the RToken.

contracts/plugins/assets/curve/stakedao/StakeDAORecursiveCollateral.sol
```diff
@@ -0,0 +1,80 @@
...
+        uint192 revenueHiding,
+        PTConfiguration memory ptConfig
+    ) CurveRecursiveCollateral(config, revenueHiding, ptConfig) {
+        IStakeDAOVault vault = IStakeDAOVault(address(config.erc20));
+        gauge = vault.liquidityGauge();
+        claimer = gauge.claimer();
+
+        uint256 rewardCount = gauge.reward_count();
+        for (uint256 i = 0; i < rewardCount; i++) {
+            rewardTokens.push(gauge.reward_tokens(i));
```
akshatmittal, line 58: AFAIK, this list can change, should move this to the `claimRewards()` function. (eg: a pool with no SDT rewards can add SDT later)

contracts/plugins/assets/curve/stakedao/StakeDAORecursiveCollateral.sol
```diff
@@ -0,0 +1,80 @@
...
+
+interface IStakeDAOGauge {
+    function claimer() external view returns (IStakeDAOClaimer);
+
+    function reward_count() external view returns (uint256);
+
+    function reward_tokens(uint256 index) external view returns (IERC20Metadata);
+}
+
+interface IStakeDAOClaimer {
```
akshatmittal, line 21: Which claimer are you looking at? The one that StakeDAO currently uses on their website has two parameters, `guages` and `claimVeSDT`. [Contract](https://etherscan.io/address/0x539e65190a371ce73244a98dec42ba635cca512c).

akshatmittal (changes requested):

contracts/plugins/assets/curve/stakedao/StakeDAORecursiveCollateral.sol
tbrent: the arg should be `false` yes?

akshatmittal (approved):

PR #1112, opened 2024-04-08
Governance as timelock canceller

tbrent:
This doesn't change functionality, just legibility. After an era change a proposal can get stuck in the timelock. This change allows it to be legibly cancelled so that dapp frontends will not continue to show it for all time.


akshatmittal:
I already made these changes in #1108, should they be in a separate PR?

julianmrodri (approved):
LGTM, added some regression tests for the different cases

PR #1113, opened 2024-04-10
Convex ETH+/ETH Collateral Plugin

tbrent:
Also in this PR: reset fork between curve collateral individual tests


contracts/plugins/assets/curve/CurveAppreciatingRTokenFiatCollateral.sol
```diff
@@ -0,0 +1,196 @@
...
+
+        // {ref/tok} = {ref/lpToken} = {lpToken@t=0/lpToken} * {1} * 2{ref/lpToken@t=0}
+        return virtualPrice.mul(rTokenRate.sqrt()).mulu(2); // LP token worth twice as much
+    }
+
+    /// @dev Warning: Can revert
+    /// @param index The index of the token: 0, 1, 2, or 3
+    /// @return low {UoA/ref_index}
+    /// @return high {UoA/ref_index}
+    function tokenPrice(uint8 index) public view override returns (uint192 low, uint192 high) {
```
akshatmittal, line 144: Needs a comment to specify that it only works for when the RToken is token0 in the pool.

akshatmittal (approved):

PR #1116, opened 2024-04-12
New arbitrum addresses

tbrent:


julianmrodri (commented):
LGTM, but we need verification in all the plugins, our wrappers,  and 3 o 4 of the core contracts (marked in the review). This would be helpful to also validate the addresses

PR #1117, opened 2024-04-15
Solidified audit mitigations

tbrent:
No mediums or highs; lows/informational only. Many mitigations optional but including anyway 

Files changed is so large mainly due to:
- using `!=` instead of `>` and `<` where possible
- `i++` -> `++i`

Other changes are included too, but these 2 are responsible for the vast majority of the diff.


contracts/p0/Broker.sol
```diff
@@ -68,7 +68,7 @@ contract BrokerP0 is ComponentP0, IBroker {
     ) external returns (ITrade) {
         assert(req.sellAmount > 0);
 
-        address caller = _msgSender();
+        address caller = msg.sender;
```
akshatmittal, line 71: `mload` is higher gas than `caller` so should not copy `msg.sender` into a local variable.

akshatmittal (approved):

contracts/p0/Broker.sol
akshatmittal: Not too important since it's p0. (I realized that later)

tbrent: Yeah I made that change everywhere in P1 but purposefully left P0 as is

pmckelvy1 (changes requested):

pmckelvy1:
lgtm

julianmrodri (commented):
The `msg.sender` change has a big impact on fuzzing. Would require restoring this everytime we merge into fuzzing as its mandatory there to be able to override the sender to simulate different senders for different txs. We should analyze if the value it provides outweights that cost

tbrent:
> LGTM. The `msg.sender` change had a big impact on fuzzing. Would require restoring this everytime we merge into fuzzing as its mandatory there to be able to override the sender to simulate different senders for different txs. Im OK leaving it for P0

I'm confused, isn't the change msg.sender -> _msgSender() in PO, and P1 remains the same?

tbrent:
> > LGTM. The `msg.sender` change had a big impact on fuzzing. Would require restoring this everytime we merge into fuzzing as its mandatory there to be able to override the sender to simulate different senders for different txs. Im OK leaving it for P0
> 
> I'm confused, isn't the change msg.sender -> _msgSender() in PO, and P1 remains the same?

Facepalm I forgot to push

tbrent:
Ah, it got clobbered. It went in at commit 4294c52d786c69794999944457519c08a367934b and then overwritten later. Fixing..

tbrent:
I have reverted the `_msgSender()` -> `msg.sender` change for the PO contracts, I did not mean to change that.

PR #1122, opened 2024-04-25
StakeDAO USDC/USDC+ plugin scripts

tbrent:
Revenue hiding amount to be confirmed by backtesting


julianmrodri (approved):
LGTM, so we dont need a wrapper in this case?

tbrent:
> LGTM, so we dont need a wrapper in this case?

I don't believe so. I believe the Gauge token is everything we need: transferrable, earns rewards, non-rebasing

tbrent:
> LGTM, so we dont need a wrapper in this case?

Don't believe so: believe the Gauge token is (i) transferrable, (ii) non-rebasing, and (iii) and earns rewards

akshatmittal (approved):
TBD: Testing on the forknet. Scripts look good!

PR #1130, opened 2024-05-01
Arbitrum convex stable plugins

julianmrodri:
* Convex `crvUSD-USDC` and `crvUSD-USDT` plugins for Arbitrum (in general for any L2)
* Does not require a wrapper. Convex Rewards Pool erc20 used directly as collateral
* Includes tests, deployment and verification scripts.


julianmrodri:
Realized we probably dont need the 2 verification scripts for USDT I added I can remove those and just leave the USDC ones and that should cover the USDT ones.

tbrent (approved):
LGTM! Really nice! 

(in the future you can skip the crvusd_usdt verification scripts: the crvusd_usdc ones will automatically apply to crvusd_usdt, so as long as you know all the collateral will be on the same chain you only need 1 verification script. it's awesome to have both around though obviously, only a plus)

julianmrodri:
> LGTM! Really nice!
> 
> (in the future you can skip the crvusd_usdt verification scripts: arbiscan will automatically apply crvusd_usdc verification to crvusd_usdt as well, so as long as you know all the collateral will be on the same chain you only need 1 verification script. it's awesome to have both around though obviously, only a plus)

I did the deployment script based on Trust's suggestion that otherwise the verification is done but in yellow state which is less trustworthy as constructor args are not validated.

tbrent:
swapping @akshatmittal in for the 2nd review

akshatmittal (approved):

PR #1140, opened 2024-05-16
Registries Setup

akshatmittal:
- [x] Version Registry
  - [x] Upgrade Changes (Main & Components)
- [x] Asset Plugin Registry
  - [x] Validations for new registrations
  - [x] Validations for existing registrations
  - [x] Optional upgrade enforcement
- [x] DAO Fee Registry
  - [x] Distributor Integration


contracts/p1/Distributor.sol
```diff
@@ -201,6 +233,33 @@ contract DistributorP1 is ComponentP1, IDistributor {
         require(rTokenDist != 0 || rsrDist != 0, "no distribution defined");
     }
 
+    function _validateDistributionTable() internal view {
+        uint256 dLen = destinations.length();
+        DAOFeeRegistry daoFeeRegistry = main.daoFeeRegistry();
+        require(destinations.length() > 0, "no destinations");
```
pmckelvy1, line 239: already have dLen

contracts/p1/Distributor.sol
```diff
@@ -201,6 +233,33 @@ contract DistributorP1 is ComponentP1, IDistributor {
...
+        uint24 daoTotals;
+
+        for (uint256 i = 0; i < dLen; ++i) {
+            address dest = destinations.at(i);
+            RevenueShare storage share = distribution[dest];
+            revTotals.rTokenTotal += share.rTokenDist;
+            revTotals.rsrTotal += share.rsrDist;
+
+            if (dest == DAO_FEE) {
+                daoTotals += share.rsrDist;
```
pmckelvy1, line 251: i think this is still an open question. would be better to just allow veRSR to set the DAO distribution entirely

contracts/p1/Distributor.sol
```diff
@@ -201,6 +233,33 @@ contract DistributorP1 is ComponentP1, IDistributor {
...
+            }
+        }
+
+        require(
```
pmckelvy1, line 255: i think there's a better way to do this.  instead of having stRSR set the distribution %'s and hoping they match what the daoFeeRegistry is expecting, we could have `_setDistribution` call the daoFeeRegistry for the % cut that it is requiring from both rsrDist and rTokenDist for this particular RToken, and then calculate what that works out to, and set the daoFee dist accordingly.

for efficiency, we can remove the singular `setDistribution` and only have 'setDistributions`, so that the final daoFee calculation can be done once everything else is set.

contracts/p1/Main.sol
```diff
@@ -58,14 +92,74 @@ contract MainP1 is Versioned, Initializable, Auth, ComponentRegistry, UUPSUpgrad
...
+
+    function upgradeRTokenTo(bytes32 versionHash, bool validateProposal) external onlyRole(OWNER) {
+        require(address(versionRegistry) != address(0), "no registry");
+        require(keccak256(abi.encodePacked(this.version())) == versionHash, "upgrade main first");
+
+        Implementations memory implementation = versionRegistry.getImplementationForVersion(
+            versionHash
+        );
+
+        if (validateProposal) {
```
pmckelvy1, line 118: can't we assume that the registry is valid before the upgrade? or is this just to cover the case where the RToken is initially upgrading to 4.0.0?

contracts/registry/DAOFeeRegistry.sol
```diff
@@ -0,0 +1,27 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+pragma solidity 0.8.19;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
+contract DAOFeeRegistry is Ownable {
```
pmckelvy1, line 6: i think there are a couple of open discussion points that we should support so as to be as generic and flexible as possible:
- fee's should be applied on a per-rtoken basis
- fee's can be applied to both rsr and rtoken revenue

contracts/registry/VersionRegistry.sol
```diff
@@ -0,0 +1,79 @@
...
+            revert VersionRegistry__InvalidRegistration();
+        }
+
+        deployments[versionHash] = deployer;
+        latestVersion = versionHash;
+
+        emit VersionRegistered(versionHash, deployer);
+    }
+
+    function deprecatedVersion(bytes32 versionHash) external onlyOwner {
```
pmckelvy1, line 47: `deprecateVersion`

pmckelvy1 (changes requested):

contracts/p1/Distributor.sol
akshatmittal: We don't want to remove a function, but we'd usually use `setDistributions`. The reason we're passing the values is because (1) they can choose to pay more than required; and (2) the calculation is gas heavy to be run at the end so better to let it be computed off-chain.

contracts/p1/Main.sol
akshatmittal: It's for _any_ upgrade including from/to 4.0.0, hence it's optional in both directions.

contracts/registry/DAOFeeRegistry.sol
akshatmittal: The fee is applied to _all_ revenue but is paid in RSR, that's how it's configured right now. We don't want the veRSR DAO to be holding RTokens.

contracts/p1/Distributor.sol
tbrent: @akshatmittal what I said earlier about not wanting to change the interface was a bad reason given this is 4.0.0

Something else we didn't consider is that under the currently-implemented design, if the DAO fee is raised, there has to be a push to get RTokens to call their `setDistributions()` functions. They may not want to, though. 


Possible other design: public() function executed at end of `setDistributions()` and externally callable only by veRSR

contracts/p1/Main.sol
```diff
@@ -49,6 +56,33 @@ contract MainP1 is Versioned, Initializable, Auth, ComponentRegistry, UUPSUpgrad
         stRSR.payoutRewards();
     }
 
+    /// Set Version Registry
+    /// @dev Can only be called once.
+    function setVerionRegistry(VersionRegistry versionRegistry_) external onlyRole(OWNER) {
```
tbrent, line 61: typo -- setVersionRegistry

contracts/p1/Main.sol
tbrent: It does seem like we need to pull apart before and after the upgrade

contracts/registry/DAOFeeRegistry.sol
tbrent: I was also thinking it would be okay to only do RSR, but I can see why we'd want to have generality

contracts/registry/VersionRegistry.sol
tbrent, line 47: typo -- `deprecateVersion()`

contracts/p1/Distributor.sol
akshatmittal: Let me rephrase, I agree with don't need to keep the same interface here, but I see no harm in keeping the old interface and also adding a new one.

I think I like the idea that veRSR distribution is _dynamic_ and always comes from the Fee Registry in real time, so it's never actually added to the distributions but rather something that's always present regardless of the entire table.

contracts/p1/Main.sol
akshatmittal: Remind me it's not a good idea to write code too late into the day.

contracts/p1/Distributor.sol
tbrent: I like this direction

pmckelvy1: yup this makes sense to me

contracts/p1/Main.sol
pmckelvy1: imo:
- required after the proposal
- optional before

contracts/p1/AssetRegistry.sol
```diff
@@ -166,6 +166,25 @@ contract AssetRegistryP1 is ComponentP1, IAssetRegistry {
...
+
+            require(
+                assetPluginRegistry.isValidAsset(
+                    keccak256(abi.encodePacked(this.version())),
+                    address(asset)
+                ),
+                "unsupported asset"
+            );
+        }
+    }
```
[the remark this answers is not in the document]
akshatmittal: I considered it, but the difference isn't much since it's basically the same loop, and this is cleaner. Lmk happy to change though.

contracts/registry/DAOFeeRegistry.sol
```diff
@@ -0,0 +1,46 @@
...
+
+    function getFeeDetails(address rToken)
+        external
+        view
+        returns (address recipient, uint256 feeNumerator)
+    {
+        recipient = feeRecipient;
+        feeNumerator = rTokenFeeNumerator[rToken] == 0
+            ? defaultFeeNumerator
+            : rTokenFeeNumerator[rToken];
```
tbrent, line 44: I believe this makes it impossible to have a 0% fee at the beginning

contracts/p1/Distributor.sol
```diff
@@ -94,24 +114,23 @@ contract DistributorP1 is ComponentP1, IDistributor {
...
-            uint256 totalShares = isRSR ? revTotals.rsrTotal : revTotals.rTokenTotal;
+            totalShares = isRSR ? revTotals.rsrTotal : revTotals.rTokenTotal;
             if (totalShares != 0) tokensPerShare = amount / totalShares;
             require(tokensPerShare != 0, "nothing to distribute");
         }
-
         // Evenly distribute revenue tokens per distribution share.
         // This rounds "early", and that's deliberate!
 
         Transfer[] memory transfers = new Transfer[](destinations.length());
```
tbrent, line 129: this needs to be `destinations.length() + 1`

contracts/p1/Distributor.sol
```diff
@@ -94,24 +114,23 @@ contract DistributorP1 is ComponentP1, IDistributor {
...
         uint256 numTransfers;
 
-        address furnaceAddr = address(furnace); // gas-saver
-        address stRSRAddr = address(stRSR); // gas-saver
```
tbrent, line 112: I assume this is for stack-too-deep reasons

contracts/p1/Distributor.sol
tbrent: but actually I think it might be simpler to keep it size `destinations.length()` and just move the DAO fee block after the transfers. don't actually need to save it since we know it's the last one

contracts/registry/DAOFeeRegistry.sol
```diff
@@ -0,0 +1,46 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+pragma solidity 0.8.19;
+
+import { Ownable } from "@openzeppelin/contracts/access/Ownable.sol";
+
+contract DAOFeeRegistry is Ownable {
+    address private feeRecipient;
+    uint256 private defaultFeeNumerator; // 1e4 = 100% fee
+    mapping(address => uint256) private rTokenFeeNumerator;
```
tbrent, line 9: let's create a `FEE_DENOMINATOR` constant?

contracts/p1/Distributor.sol
```diff
@@ -160,6 +194,13 @@ contract DistributorP1 is ComponentP1, IDistributor {
             revTotals.rTokenTotal += share.rTokenDist;
             revTotals.rsrTotal += share.rsrDist;
         }
+
+        // DAO Fee
+        (, uint256 feeNumerator) = main.daoFeeRegistry().getFeeDetails(address(rToken));
+        revTotals.rsrTotal += uint24(
+            (feeNumerator * uint256(revTotals.rTokenTotal + revTotals.rsrTotal)) /
+                (1e4 - feeNumerator)
```
tbrent, line 202: should read the 1e4 out from `daoFeeRegistry()`, or could add it as a return value to `getFeeDetails()`. but shouldn't just in-line the 1e4 imo

contracts/registry/DAOFeeRegistry.sol
tbrent: or return it in `getFeeDetails()`

contracts/registry/DAOFeeRegistry.sol
akshatmittal: Not impossible right, since defaultFee can be set to 0?

contracts/p1/Distributor.sol
akshatmittal: Yep

contracts/registry/DAOFeeRegistry.sol
akshatmittal: Yep, wanted to do that but I see that we use this 1e4 pattern elsewhere in the protocol so didn't do it assuming it was documented globally somewhere.

akshatmittal: Nothing specifically that this does assume that the same 1e4 denominator is used everywhere for revenue purposes.

contracts/registry/DAOFeeRegistry.sol
tbrent: Sure -- but wouldn't be able to set individual RToken fees to 0 if the default were nonzero

tbrent: Might be okay

contracts/registry/DAOFeeRegistry.sol
tbrent: right -- forgot we already had a constant for it in `IDistributor`: `MAX_DISTRIBUTION`

contracts/registry/DAOFeeRegistry.sol
akshatmittal: Fairly easy to add a "isFeeSet" mapping so lmk

tbrent:
New protocol requirement being introduced with this PR: distributor table must sum to beyond 1e4. This is to ensure the DAO fee has enough precision.

contracts/p1/AssetRegistry.sol
```diff
@@ -166,6 +166,25 @@ contract AssetRegistryP1 is ComponentP1, IAssetRegistry {
         }
     }
 
+    /// @inheritdoc IAssetRegistry
+    function validateCurrentAssets() external view {
```
julianmrodri, line 170: We should add a check here to revert elegantly if the registries are not set up. Otherwise it ends up reverting with a " fuinction returns unexpected amount of data" when trying to return the value for the first item, whent he Asset Plugin Registry address is 0 for example

julianmrodri (changes requested):

test/Main.test.ts
```diff
@@ -1167,6 +1167,10 @@ describe(`MainP${IMPLEMENTATION} contract`, () => {
       ).to.be.revertedWith('not enough gas to unregister safely')
     })
 
+    it('Should validate current assets if no Plugin Registry', async () => {
+      expect(await assetRegistry.validateCurrentAssets()).to.not.be.reverted
```
akshatmittal, line 1171: Shouldn't this be `await expect(assetRegistry...` or am I confusing this?

PR #1141, opened 2024-05-16
ETH+/ETH collateral feedback

tbrent:


contracts/plugins/assets/curve/CurveRecursiveCollateral.sol
```diff
@@ -83,25 +85,42 @@ contract CurveRecursiveCollateral is CurveStableCollateral {
...
+
+            // === Update exposedReferencePrice, ignoring default ===
+
+            // {ref/tok} = {ref/tok} * {1}
+            uint192 hiddenReferencePrice = underlyingRefPerTok_.mul(revenueShowing);
+
+            // uint192(<) is equivalent to Fix.lt
+            if (underlyingRefPerTok_ < exposedReferencePrice) {
+                exposedReferencePrice = underlyingRefPerTok_;
+                // markStatus(CollateralStatus.DISABLED); // don't DISABLE
```
akshatmittal, line 117: If we don't intend to default based on this, why don't we just ignore this whole revenue hiding for the reference price and pass it as-is?

test/plugins/individual-collateral/curve/cvx/CvxAppreciatingRTokenSelfReferential.test.ts
```diff
@@ -340,6 +340,23 @@ const collateralSpecificStatusTests = () => {
     await expect(collateral.refresh()).to.emit(collateral, 'CollateralStatusChanged').withArgs(1, 0)
     expect(await collateral.status()).to.equal(0)
   })
+
+  it('Read-only reentrancy', async () => {
```
akshatmittal, line 344: Very cool to see this test!

akshatmittal (changes requested):

contracts/plugins/assets/curve/CurveRecursiveCollateral.sol
tbrent: Because the core contracts don't want refPerTok() to decrease. If refPerTok() oscillates tightly it can result in seizing RSR, processing revenue, seizing RSR, etc.

tbrent: each time that happens the RToken holders make out like bandits and RSR stakers get rekt

tbrent: refPerTok() should only ever decrease after the inner RToken takes a haircut to basketsNeeded()

akshatmittal: Oh yeah, this is not the plugin where we're overriding that. Got it, still relies on the exposedReferencePrice for the refPerTok external output.

akshatmittal (approved):

PR #1142, opened 2024-05-17
arbitrum crvusd crvusdt deployments

julianmrodri:
* register deployments on Arbitrum for the `crvUSD-USDC` and `crvUSD-USDT` plugins


tbrent (approved):
LGTM. 

One thing I noticed while going through this is that the deployment scripts do `defaultThreshold: DEFAULT_THRESHOLD` for a hardcoded `DEFAULT_THRESHOLD`, but I don't think this is enough to merit any sort of re-deployment or fix. Looks like the hard-code was to 2%: our rule would normally be 1% plus the combined oracle error, which I think comes to like 1.7% anyway. Should be fine.

julianmrodri:
OK. Its not expensive to redeploy we can do that. I'll fix it in the script

PR #1144, opened 2024-05-18
First try at adding a number of impact assets for use as collateral.

TerexitariusStomp:
We need more stablecoins that use impact assets (renewable energy, regenerative agriculture, etc.) as collateral. This is my first attempt to add them to the Reserve Protocol.

### DIFF
--- a/contracts/plugins/assets/impactassets/BRKLCollateral.sol
+++ b/contracts/plugins/assets/impactassets/BRKLCollateral.sol
@@ -1,0 +1,66 @@
+// SPDX-License-Identifier: BlueOak-1.0.0
+pragma solidity 0.8.19;
+
+import "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
+import "../../libraries/Fixed.sol";
+import "./FiatCollateral.sol";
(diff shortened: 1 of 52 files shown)


pmckelvy1:
thanks for the submissions @TerexitariusStomp .

there is a lot more work here in order for a collateral to be able to be used by RTokens (like proper contract address, and pricing oracles).  please see [this guide](https://github.com/reserve-protocol/protocol/blob/master/docs/writing-collateral-plugins.md) on how to write collateral plugins.  it has links to other docs that go into more detail about how they function within the protocol.

also feel free to reach out in the [#technical-discussions](https://discord.com/channels/920041058941956097/1070255658768814142) channel in the Reserve Protocol discord if you need answers to more direct questions.

closing this pr for now.

TerexitariusStomp:
@tbrent and @pmckelvy1 thanks for your review. I'm still learning to code, will review the documents and try again. Out of curiosity would either of you be willing to be paid to create these plugins? If so would love to know a quote. I've been in contact with oracle providers such as API3 and they're wanting around 2k a month for the data stream per token.Energy Web and Powerledger are the only tokens with enough market cap for API3 to handle. It seems that new stablecoins issued with Reserve pay for their oracle and this doens't come from the Reserve treasury, is this correct?

PR #1150, opened 2024-05-30
fix GnosisTrade approval

tbrent:


akshatmittal (approved):

contracts/plugins/trading/GnosisTrade.sol
```diff
@@ -139,7 +139,7 @@ contract GnosisTrade is ITrade, Versioned {
         //
         // Context: wcUSDCv3 has a non-standard approve() function that reverts if the approve
         // amount is > 0 and < type(uint256).max.
-        AllowanceLib.safeApproveFallbackToMax(address(sell), address(gnosis), initBal);
+        AllowanceLib.safeApproveFallbackToMax(address(sell), address(gnosis), req.sellAmount);
```
akshatmittal, line 142: Just want to confirm Gnosis behaviour, we _don't_ want to use `_sellAmount` here because fee, if set, are paid from `sellAmount`.

tbrent: Yes, exactly.

tbrent:
looks like the CI failures here are from the base branch, so i'm going ahead with merging given the simplicity of the change

PR #1153, opened 2024-06-05
Gnosis trade approval pt2

tbrent:
The first PR #1150 missed these two other places we needed to use the `req.sellAmount` instead of the `initBal`


tbrent:
The `Broker` checks to see it doesn't send a `GnosisTrade` more tokens than it can deal with, but an attacker can always donate tokens to drive the `initBal` above uint96.

akshatmittal (approved):

PR #1154, opened 2024-06-05
Extreme tests 21 and 27 decimals

julianmrodri:


tbrent:
Someone else should review this PR since I did the implementation changes.

Roughly: went through all `shiftl_toUint()` calls and added specific rounding. This was almost entirely FLOOR but ended up being CEIL in one or two places. Also there's new math in GnosisTrade to check.

akshatmittal (approved):
Approving with the caveat that we'll do at least a couple audits on this codebase, with special considerations for this.

PR #1157, opened 2024-06-13
Fix Rtoken Exteme test

julianmrodri:
* Scale tokens to the appropiate decimals when minting in test


julianmrodri:
Merging after offline approval

PR #1163, opened 2024-06-30
DRAFT: Feat/on chain redemptions

jankjr:
I'm not entirely sure where this would fit in so i'm pushing it here for now.

This script will redeem ETH+ directly for ETH and follows an interview that mimicks uniswap v2. It only works for ETH+ into ETH, it actually performs relatively well for smaller amounts at least ignoring gas costs.

Perhaps this could make a sort of routers part of the `Facade`, and add a few predefined and maintained ways to exit an RToken position into a single underlying? @tbrent


tbrent:
I think that's exactly the right approach -- we can deploy this contract and hook our Facade to point to it. I'm going to hold off until the rETH-trading case is in the contract but this sounds like the plan for after.

PR #1173, opened 2024-07-12
issuance premium

tbrent:
Add issuance premium to BasketHandler that begins on and can be toggled off by governance.

One of the bigger things to note is that this changes the price of a BU pretty significantly, since the upper BU price now takes into account any issuance premium that may apply. Interestingly this means during a de-peg the BU high price remains at $1.01, or just above its peg, instead of falling with the low price. This part should get some special attention, it feels more potentially dangerous to me than the actual issuance quantity change. 

Had to do a few things to reduce BasketHandler contract size as well, such as finally deleting `lotPrice()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@ TODO
 
 Make sure distributor table sums to >10000.
 
+Upgrade ALL collateral plugins
+
 ## Core Protocol Contracts
 
 - `AssetRegistry`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This release prepares the core protocol for veRSR through the introduction of 3 
 
 The release also expands collateral decimal support from 18 to 21, with some caveats about minimum token value. See [docs/solidity-style.md](./docs/solidity-style.md#Collateral-decimals) for more details.
 
+Finally, it adds resistance to toxic issuance by charging more when the collateral is under peg. This has the side-effect of also creating a new revenue source.
+
 ## Upgrade Steps
 
 TODO
@@ -15,10 +17,14 @@ Make sure distributor table sums to >10000.
 ## Core Protocol Contracts
 
 - `AssetRegistry`
+
   - Prevent registering assets that are not in the `AssetPluginRegistry`
   - Add `validateCurrentAssets() view`
+
 - `BackingManager`
   - Switch from sizing trades using the low price to the high price
+- `BasketHandler`
+  - Increase `quote()` quantities during issuance when collateral is under peg
 - `Broker`
   - Make setters only callable by `Main`
 - `Distributor`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ Upgrade ALL collateral plugins
 - `BackingManager`
   - Switch from sizing trades using the low price to the high price
 - `BasketHandler`
-  - Increase `quote()` quantities during issuance to include an issuance premium for de-pegged tokens
+  - Increase `quote()` quantities during issuance to include an issuance premium for de-pegged tokens, up to 50%
   - Remove `lotPrice()`
 - `Broker`
   - Make setters only callable by `Main`
@@ -47,6 +47,7 @@ Upgrade ALL collateral plugins
 - Support expanded from 18 to 21 decimals, with minimum collateral token value requirement of `$0.001` at-peg.
 - FLOOR rounding added explicitly to `shiftl_toFix`
 - Remove lotPrice()
+- Set a minimum default threshold of 70%
 
 ### Trading
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,6 +48,10 @@ Make sure distributor table sums to >10000.
 - Remove lotPrice()
 - Set a minimum default threshold of 70%
 
+#### Collateral
+
+Add `savedPegPrice` to `ICollateral` interface
+
 ### Trading
 
 - `GnosisTrade`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ Upgrade ALL collateral plugins
 - `BackingManager`
   - Switch from sizing trades using the low price to the high price
 - `BasketHandler`
-  - Increase `quote()` quantities during issuance when collateral is under peg
+  - Increase `quote()` quantities during issuance to include an issuance premium for de-pegged tokens
   - Remove `lotPrice()`
 - `Broker`
   - Make setters only callable by `Main`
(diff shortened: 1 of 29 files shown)


tbrent:
issuance premiums for all the currently-backing collateral

Screenshot 2024-07-13 at 8 55 58 PM

tbrent:
depegs

tbrent:
still-to-add: a single fork test case on the new functionality

CHANGELOG.md
```diff
@@ -6,19 +6,27 @@ This release prepares the core protocol for veRSR through the introduction of 3
...
 ## Core Protocol Contracts
 
 - `AssetRegistry`
   - Prevent registering assets that are not in the `AssetPluginRegistry`
   - Add `validateCurrentAssets() view`
 - `BackingManager`
   - Switch from sizing trades using the low price to the high price
+- `BasketHandler`
+  - Increase `quote()` quantities during issuance to include an issuance premium for de-pegged tokens
+  - Add `setSkipIssunancePremium()`, callable by governance
```
akshatmittal, line 28: The setter is a double negative (Also a typo), ideally use something like `setIssuancePremiumEnabled` instead.

contracts/p1/BasketHandler.sol
```diff
@@ -339,19 +340,21 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
             (block.timestamp >= lastStatusTimestamp + warmupPeriod);
     }
 
+    /// Basket quantity rounded up, without any issuance premium
     /// @param erc20 The token contract to check for quantity for
     /// @return {tok/BU} The token-quantity of an ERC20 token in the basket.
     // Returns 0 if erc20 is not registered or not in the basket
     // Returns FIX_MAX (in lieu of +infinity) if Collateral.refPerTok() is 0.
     // Otherwise returns (token's basket.refAmts / token's Collateral.refPerTok())
     function quantity(IERC20 erc20) public view returns (uint192) {
```
akshatmittal, line 349: Noting that all external functions now do not expose the premium, so maybe worth adding another getter here to allow passing issuancePremium flag.

contracts/plugins/assets/curve/CurveStableMetapoolCollateral.sol
```diff
@@ -122,7 +122,8 @@ contract CurveStableMetapoolCollateral is CurveStableCollateral {
         high = aumHigh.div(supply, CEIL);
         assert(low <= high); // not obviously true just by inspection
 
-        return (low, high, 0);
+        pegPrice = 0; // can't deduce from MEV-manipulable pricing unfortunately
```
akshatmittal, line 125: We should make it clear that 0 is a special state where the BasketHandler will skip adding the issuance premium if the value is 0.

contracts/p1/BasketHandler.sol
```diff
@@ -361,70 +364,72 @@ contract BasketHandlerP1 is ComponentP1, IBasketHandler {
...
 
-    /// Should not revert
-    /// lowLow should be nonzero when the asset might be worth selling
-    /// @dev Deprecated. Phased out in 3.1.0, but left on interface for backwards compatibility
-    /// @return lotLow {UoA/BU} The lower end of the lot price estimate
-    /// @return lotHigh {UoA/BU} The upper end of the lot price estimate
-    // returns sum(quantity(erc20) * lotPrice(erc20) for erc20 in basket.erc20s)
-    function lotPrice() external view returns (uint192 lotLow, uint192 lotHigh) {
-        return _price(true);
+            uint192 pegPrice = coll.savedPegPrice(); // {target/ref}
```
akshatmittal, line 394: This is a significant breaking change right, `savedPegPrice` doesn't exist in any existing collateral so all plugins MUST be upgraded to a new version when upgrading.

akshatmittal (commented):
Core Protocol changes look good, briefly looked at the plugins as well and they just make them complete instead of adding new functionality so it seems good. Noting that this will make this release a breaking change even beyond what we've been previously discussing with new _required_ getters on the plugin.

CHANGELOG.md
tbrent: That's fair -- I want the variable to keep the signage it has but I'll change the setter and event

contracts/p1/BasketHandler.sol
tbrent: Not quite -- `quote(, CEIL)` and `price()` expose the premium. Agreed no `quantity*()` functions do though. I don't think we need the flag.

tbrent: The BackingManager has a call to `quantityUnsafe()` that _shouldn't_ use the flag. If it makes sense to add the flag I think it's because something outside our system needs it. So the question is whether `quote()` is sufficient.

contracts/p1/BasketHandler.sol
tbrent: Corrrrrect. I was initially thinking this would be okay because we'd just deploy 4.0.0 versions across the board, but with 3.4.0 there were a few cases of collateral left behind on <3.4.0 versions. 

Options seem to be try-catch or make sure to have 4.0.0 versions of everything.

contracts/p1/BasketHandler.sol
akshatmittal: I think it's worth adding another function, purely because an external system might need to know the premium on the actual basket instead of just a quote even though both can be derived from each other.

contracts/p1/BasketHandler.sol
tbrent: It's too bad this isn't a fit for the AssetPluginRegistry...it's no use for already-registered assets (on purpose)

contracts/p1/BasketHandler.sol
tbrent: There isn't a premium at the basket level, conceptually. Individual collateral can be required more of, but you can't combine the quantities of the different tokens without something like applying a USD lens. So it doesn't make sense to me to expose premium at a basket level. 

`quote()` seems to offload _more_ work for integrators here, no? You can directly divide the quantities for the `quote(FLOOR)` and `quote(CEIL)` baskets to see how much premium is being applied.

tbrent: Trying to steelman...I guess is someone is writing an on-chain integration that only cares about one collateral, then they'd want individual collateral level access? Mmm I think I see what to do here...

tbrent: Trying to steelman...I guess is someone is writing an on-chain integration that only cares about one collateral, then they'd want individual collateral level access? Mmm I think I see what to do here...

tbrent: ok, added a getter for what I think they would care about. just under the contract size!

contracts/p1/BasketHandler.sol
tbrent: Went with try-catch

PR #1174, opened 2024-07-17
ETHx integration

dulguun-staderlabs:


contracts/plugins/assets/ethx/README.md
```diff
@@ -0,0 +1,26 @@
...
+### Units
+
+| tok  | ref  | target | UoA |
+| ---- | ---- | ------ | --- |
+| ETHx | ETH2 | ETH    | USD |
+
+### refPerTok()
+
+Gets the exchange rate for `ETHx` to `ETH2` from the ETHx token contract using the [getExchangeRate()]()
+function. This is the rate used by rocket pool when converting between ethx and eth2 and is closely followed by secondary markets.
```
blockgroot, line 18:
```suggestion
function. This is the rate used by stader labs when converting between ethx and eth2 and is closely followed by secondary markets.
```

julianmrodri:
I noticed you are using `getExchangeRate()` for the `underlyingRefPerTok()` calculation. But that function does not exist on the `ETHx` token. How can you get the equivalent for that in Stader Ethx? Its the amount of ETH2 per each ETHx.

dulguun-staderlabs:
> @dulguun-staderlabs I noticed you are using `getExchangeRate()` for the `underlyingRefPerTok()` calculation. But that function does not exist on the `ETHx` token. How can you get the equivalent for that in Stader Ethx? Its the amount of underlying ETH2 per each ETHx. WE should be able to obtain this rate from your protocol ideally.

That is correct.
Actual contract is called `StaderStakePoolsManager` at `0xcf5EA1b38380f6aF39068375516Daf40Ed70D299` (eth mainnet).It has a function `function getExchangeRate() external view override returns (uint256)`.

I was thinking of using `ETHx` as simplicity but can definitely change it to the actual contract name.

julianmrodri:
> > @dulguun-staderlabs I noticed you are using `getExchangeRate()` for the `underlyingRefPerTok()` calculation. But that function does not exist on the `ETHx` token. How can you get the equivalent for that in Stader Ethx? Its the amount of underlying ETH2 per each ETHx. WE should be able to obtain this rate from your protocol ideally.
> 
> That is correct. Actual contract is called `StaderStakePoolsManager` at `0xcf5EA1b38380f6aF39068375516Daf40Ed70D299` (eth mainnet).It has a function `function getExchangeRate() external view override returns (uint256)`.
> 
> I was thinking of using `ETHx` as simplicity but can definitely change it to the actual contract name.

So what the plugin is missing is the link to that contract and that is the one that should be used in `underlyingRefPerTok()` instead of using the `erc20` variable. Is there a way to get to that address fron the `ETHx` contract? we can get it dynamically in the constructor and set it as `immutable`

dulguun-staderlabs:
> > > @dulguun-staderlabs I noticed you are using `getExchangeRate()` for the `underlyingRefPerTok()` calculation. But that function does not exist on the `ETHx` token. How can you get the equivalent for that in Stader Ethx? Its the amount of underlying ETH2 per each ETHx. WE should be able to obtain this rate from your protocol ideally.
> > 
> > 
> > That is correct. Actual contract is called `StaderStakePoolsManager` at `0xcf5EA1b38380f6aF39068375516Daf40Ed70D299` (eth mainnet).It has a function `function getExchangeRate() external view override returns (uint256)`.
> > I was thinking of using `ETHx` as simplicity but can definitely change it to the actual contract name.
> 
> So what the plugin is missing is the link to that contract and that is the one that should be used in `underlyingRefPerTok()` instead of using the `erc20` variable. Is there a way to get to that address fron the `ETHx` contract? we can get it dynamically in the constructor and set it as an `immutable` var. Then use that contract in the `underlyingRefPerTok()` calculation.
> 
> I would use the correct naming for the variable and interface (instead of IETHx) (in the vendor folder for example and in the plugin)

I changed accordingly. ✅

julianmrodri:
Pushed some refactors and the tests.. We need to open this PR against the `ethx-plugin` branch. I can merge it there and take the next steps.

PR #1180, opened 2024-07-30
Backing buffer facet

tbrent:
Results from running this on mainnet on the largest 3 RTokens

```
- ETH+: $330k required buffer; has $337k
- USD3: $35k required buffer; has $39k
- eUSD: $2,300 required buffer; has $7500
```

(makes sense, because recall eUSD buffer is set too low atm)


julianmrodri:
LGTM, fixed a small typo

julianmrodri (approved):

PR #1185, opened 2024-08-19
Lower default thresholds

julianmrodri:
* Lower default thresholds when relying on single price feed (avoid using combined error)


scripts/deployment/phase2-assets/collaterals/deploy_cbeth_collateral.ts
```diff
@@ -64,7 +65,7 @@ async function main() {
         maxTradeVolume: fp('1e6').toString(), // $1m,
         oracleTimeout: '3600', // 1 hr,
         targetName: hre.ethers.utils.formatBytes32String('ETH'),
-        defaultThreshold: fp('0.02').add(oracleError).toString(), // ~4.5%
+        defaultThreshold: fp('0.02').add(CBETH_ORACLE_ERROR).toString(), // 3%
```
tbrent, line 68: I think cbETH should use the combined oracle error, I'm seeing that the two oracle prices get multiplied together.

scripts/deployment/phase2-assets/collaterals/deploy_cbeth_collateral.ts
```diff
@@ -91,7 +93,7 @@ async function main() {
         maxTradeVolume: fp('1e6').toString(), // $1m,
         oracleTimeout: '1200', // 20 min
         targetName: hre.ethers.utils.formatBytes32String('ETH'),
-        defaultThreshold: fp('0.02').add(oracleError).toString(), // ~2.5%
+        defaultThreshold: fp('0.02').add(CBETH_ORACLE_ERROR).toString(), // 2.5%
```
tbrent, line 96: cbETH

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
```diff
@@ -181,7 +181,7 @@ async function main() {
         maxTradeVolume: fp('1e6'), // $1m,
         oracleTimeout: '86400', // 24 hr
         targetName: ethers.utils.formatBytes32String('BTC'),
-        defaultThreshold: fp('0.01').add(combinedBTCWBTCError), // ~3.5%
+        defaultThreshold: fp('0.01').add(wbtcOracleError), // 3%
```
tbrent, line 184: same thing here, I'm seeing in `MorphoNonFiatCollateral` that the two prices are multiplied together

tbrent (changes requested):
I stopped reviewing halfway through because I think it's correct for these to use the combined oracle error.

scripts/deployment/phase2-assets/collaterals/deploy_cbeth_collateral.ts
julianmrodri: It uses the two oracles for `low` and `high`, but only one for `pegPrice` which is what we use for `defaultThreshold`. Should we still consider both oracles in the default threshold calculation?

scripts/deployment/phase2-assets/collaterals/deploy_morpho_aavev2_plugin.ts
julianmrodri: Same thing as for `cbETH`.. `pegPrice` uses a single oracle.

scripts/deployment/phase2-assets/collaterals/deploy_cbeth_collateral.ts
tbrent: Ah got it, my bad

tbrent (approved):
LGTM

PR #1190, opened 2024-09-04
Deploy pyusd

pmckelvy1:


pmckelvy1:
https://etherscan.io/address/0x7622d205b88e5fb7030e5d717ade7202592b31ab#code

hardhat.config.ts
```diff
@@ -81,9 +83,7 @@ const config: HardhatUserConfig = {
     mainnet: {
       chainId: 1,
       url: MAINNET_RPC_URL,
-      accounts: {
-        mnemonic: MNEMONIC,
-      },
+      accounts: [DEPLOYER_PK],
```
akshatmittal, line 86: Do not use PK, use the mnemonic setup instead. Several scripts in the project depend on having multiple available accounts, and generally speaking it's better anyway.

scripts/verification/assets/verify_pyusd.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  if (developmentChains.includes(hre.network.name)) {
+    throw new Error(`Cannot verify contracts for development chain ${hre.network.name}`)
+  }
+
+  deployments = <IAssetCollDeployments>getDeploymentFile(getAssetCollDeploymentFilename(chainId))
+
+  const asset = await hre.ethers.getContractAt('Asset', deployments.assets.pyUSD!)
+
+  /** ******************** Verify RSR Asset ****************************************/
```
tbrent, line 26: comment

hardhat.config.ts
tbrent, line 86: why add this DEPLOYER_PK approach? trying to separate out keys used for different purposes?

tbrent: maybe you didn't mean to check this in?

tbrent:
deployed address LGTM 👍

hardhat.config.ts
pmckelvy1: yeah i generally prefer to use a pk, as it is less surface area for moving sensitive info around. if i accidentally check in the seed then that entire wallet is compromised (not that i'm putting anything significant on these addresses anyways.
i'll roll it back

scripts/deploy.ts
```diff
@@ -88,7 +88,8 @@ async function main() {
       'phase2-assets/collaterals/deploy_apxeth.ts',
       'phase2-assets/collaterals/deploy_USDe.ts',
       'phase2-assets/assets/deploy_crv.ts',
-      'phase2-assets/assets/deploy_cvx.ts'
+      'phase2-assets/assets/deploy_cvx.ts',
+      'phase2-assets/assets/deploy_pyUSD.ts'
```
tbrent, line 92: case mismatch with file name

tbrent (approved):
nit: update `scripts/verify_etherscan.ts` too, LGTM after

tbrent (approved):

akshatmittal (approved):

PR #1209, opened 2024-09-19
Require redemption throttle limits above issuance throttle limits

pmckelvy1:


contracts/p1/RToken.sol
```diff
@@ -79,8 +80,8 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
         furnace = main_.furnace();
 
         mandate = mandate_;
-        setIssuanceThrottleParams(issuanceThrottleParams_);
         setRedemptionThrottleParams(redemptionThrottleParams_);
+        setIssuanceThrottleParams(issuanceThrottleParams_);
```
tbrent, line 84: calling the hybrid setter here seems slightly better. would have to make `public`

contracts/p1/RToken.sol
```diff
@@ -443,6 +444,38 @@ contract RTokenP1 is ComponentP1, ERC20PermitUpgradeable, IRToken {
...
+        _setRedemptionThrottleParams(redemptionParams);
+        require(
+            isRedemptionThrottleGreaterByDelta(issuanceParams, redemptionParams),
+            "redemption throttle too low"
+        );
+    }
+
+    // === Private Helpers ===
+
+    function _setIssuanceThrottleParams(ThrottleLib.Params calldata params) public governance {
```
tbrent, line 478: duplicate `governance` modifiers on the internal functions. I don't have much of a preference for which to keep, I guess if I _had_ to pick I would go external

tbrent (approved):

PR #1210, opened 2024-09-20
Aerodrome Collateral Plugin (Base)

akshatmittal:
Aerodrome Stable Pools plugin


tbrent (approved):
LGTM

akshatmittal:
LGTM

Should probably add a comment somewhere about gauge selection in this plugin, since there can be multiple gauges for Aerodrome pools. To extend that, probably add a note somewhere that if a new gauge is deployed for the associated pool, a new wrapper will have to be registered and collateral switch, etc.

pmckelvy1:
lgtm

PR #1212, opened 2024-09-24
tweak oracle factory to also support chainlink interface

tbrent:


contracts/facade/factories/ExchangeRateOracleFactory.sol
```diff
@@ -0,0 +1,79 @@
...
+        view
+        returns (
+            uint80 roundId,
+            int256 answer,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
+        )
+    {
+        return (0, int256(exchangeRate()), 0, 0, 0);
```
akshatmittal, line 45: I don't think `startedAt` and `updatedAt` can be zero here, but not sure if Morpho requires that (could use `block.timestamp` for both?). I'd also set the `roundId` and `answeredInRound` to `1` instead of `0`.

tbrent:
closing in favor of #1222

PR #1214, opened 2024-09-30
Morpho eUSD Plugin (Base)

julianmrodri:
Metamorhp Morpho EUSD Pluing on Base


scripts/deployment/phase2-assets/collaterals/deploy_morphoeUSD.ts
```diff
@@ -0,0 +1,94 @@
...
+    ).deploy(
+      {
+        priceTimeout: PRICE_TIMEOUT.toString(),
+        chainlinkFeed: eUSD_USD_FEED,
+        oracleError: eUSD_ORACLE_ERROR.toString(),
+        erc20: networkConfig[chainId].tokens.meUSD,
+        maxTradeVolume: fp('1e6').toString(), // 17m vault
+        oracleTimeout: eUSD_ORACLE_TIMEOUT.toString(),
+        targetName: hre.ethers.utils.formatBytes32String('USD'),
+        defaultThreshold: eUSD_ORACLE_ERROR.add(fp('0.01')).toString(), // +1% buffer rule
```
pmckelvy1, line 68: do we need to add a full 1% here? we're targetting base, so the oracles should be quick to update, and the base oracle error is 0.5%. feels like we could get away with 1% total? (oracle error + 0.5%). @tbrent thoughts?

scripts/deployment/phase2-assets/collaterals/deploy_morphoeUSD.ts
```diff
@@ -0,0 +1,94 @@
...
+        delayUntilDefault: DELAY_UNTIL_DEFAULT.toString(),
+      },
+      fp('1e-4') // can have mild drawdowns
```
pmckelvy1, line 71: morpho drawdowns are this big?!

tbrent: We actually used 10x this on mainnet for Re7WETH after backtesting and finding it defaulted below that:

https://github.com/reserve-protocol/protocol/blob/14da963751b59faff6fb6081fcbdcfe8b99e9b2a/scripts/deployment/phase2-assets/collaterals/deploy_re7weth.ts#L68

scripts/deployment/phase2-assets/collaterals/deploy_morphoeUSD.ts
tbrent: The +1% rule doesn't interact with the attributes of the oracle in my mind, eg oracleError or oracleTimeout, etc. 

Anyway, the oracleTimeout is 24h. I don't see why it being on base would change anything, this is still the value that governs how often they will push heartbeat updates.

scripts/deployment/phase2-assets/collaterals/deploy_morphoeUSD.ts
tbrent: 🤞 hoping `1e-4` is enough for eUSD

pmckelvy1: i think we have backtesting on base? ill see if we can run this

pmckelvy1 (approved):
lgtm

PR #1215, opened 2024-10-02
feat: add USDA plugin and scripts

nlecoufl:


common/configuration.ts
```diff
@@ -513,6 +519,8 @@ export const networkConfig: { [key: string]: INetworkConfig } = {
       sUSDbC: '0x4c80e24119cfb836cdf0a6b53dc23f04f7e652ca',
       wstETH: '0xc1CBa3fCea344f92D9239c08C0568f6F2F0ee452',
       STG: '0xE3B53AF74a4BF62Ae5511055290838050bf764Df',
+      USDA: '0x0000206329b97DB379d5E1Bf586BbDB969C63274',
```
pmckelvy1, line 522: just noting that basescan shows `agUSD` and not `USDA`. is this on purpose?

scripts/addresses/1-tmp-assets-collateral.json
```diff
@@ -61,7 +61,8 @@
     "sfrxETH": "0x4c891fCa6319d492866672E3D2AfdAAA5bDcfF67",
     "apxETH": "0x05ffDaAA2aF48e1De1CE34d633db018a28e3B3F5",
     "sUSDe": "0x35081Ca24319835e5f759163F7e75eaB753e0b7E",
-    "pyUSD": "0xa5cde4fB1132daF8f4a0D3140859271208d944E9"
+    "pyUSD": "0xa5cde4fB1132daF8f4a0D3140859271208d944E9",
+    "stUSD": "0xb6aD2903fAC2db628624DeE61e217F50Ad54638E"
```
pmckelvy1, line 65: i don't see anything deployed here https://etherscan.io/address/0xb6aD2903fAC2db628624DeE61e217F50Ad54638E

test/plugins/individual-collateral/angle/USDAFiatCollateral.test.ts
```diff
@@ -0,0 +1,223 @@
...
+  targetName: ethers.utils.formatBytes32String('USD'),
+  rewardERC20: ZERO_ADDRESS,
+  priceTimeout: PRICE_TIMEOUT,
+  chainlinkFeed: USDA_USD_PRICE_FEED,
+  oracleTimeout: ORACLE_TIMEOUT,
+  oracleError: ORACLE_ERROR,
+  maxTradeVolume: MAX_TRADE_VOL,
+  defaultThreshold: DEFAULT_THRESHOLD, // 72 hs
+  delayUntilDefault: DELAY_UNTIL_DEFAULT,
+  revenueHiding: fp('0'),
```
pmckelvy1, line 49: `1e-6` to match the deployment scripts

contracts/plugins/assets/angle/USDAFiatCollateral.sol
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @title USDA Fiat Collateral
+ * @notice Collateral plugin for USDA (Angle)
+ * tok = stUSD (ERC4626 vault)
+ * ref = USDA
+ * tar = USD
+ * UoA = USD
+ */
+
+contract USDAFiatCollateral is ERC4626FiatCollateral {
```
pmckelvy1, line 16: `stUSDACollateral` or `StakedUSDACollateral` seem like more descript names, since the collateral is stUSDA and not USDA

contracts/plugins/assets/angle/USDAFiatCollateral.sol
```diff
@@ -0,0 +1,25 @@
...
+    /// config.erc20 must be stUSD
+    /// @param config.chainlinkFeed Feed units: {UoA/ref}
+    /// @param revenueHiding {1} A value like 1e-6 that represents the maximum refPerTok to hide
+    constructor(CollateralConfig memory config, uint192 revenueHiding)
+        ERC4626FiatCollateral(config, revenueHiding)
+    {
+        require(config.defaultThreshold != 0, "defaultThreshold zero");
+    }
+}
```
pmckelvy1, line 25: looking at the stUSDA ERC4626 contract, and there seems to be an internal accrual happening on any deposit or withdrawal. since this accrual affects the pricePerShare() rate, we should make sure this gets updated on any call to `collateral.refresh()` (refresh() ensures that all state variables affecting price are updated so that the most accurate, up-to-date price can be calculated).

the best way to do this is probably to copy over the refresh call from the `AppreciatingFiatCollateral` and add a call to `deposit(0, address(this))` at the top of the function.

scripts/verification/collateral-plugins/verify_usda.ts
```diff
@@ -0,0 +1,59 @@
...
+  deployments = <IAssetCollDeployments>getDeploymentFile(assetCollDeploymentFilename)
+
+  /********  Verify stUSD COllateral  **************************/
+  await verifyContract(
+    chainId,
+    deployments.collateral.stUSD,
+    [
+      {
+        priceTimeout: PRICE_TIMEOUT.toString(),
+        chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDC,
```
pmckelvy1, line 41: USDA

scripts/deployment/phase2-assets/collaterals/deploy_usda.ts
```diff
@@ -0,0 +1,92 @@
...
+  let collateral: USDAFiatCollateral
+
+  const USDAFiatCollateralFactory: ContractFactory = await hre.ethers.getContractFactory(
+    'USDAFiatCollateral'
+  )
+
+  collateral = <USDAFiatCollateral>await USDAFiatCollateralFactory.connect(deployer).deploy(
+    {
+      priceTimeout: PRICE_TIMEOUT.toString(),
+      chainlinkFeed: networkConfig[chainId].chainlinkFeeds.USDC,
```
pmckelvy1, line 58: USDA

test/plugins/individual-collateral/angle/constants.ts
```diff
@@ -0,0 +1,17 @@
+import { bn, fp } from '../../../../common/numbers'
+import { networkConfig } from '../../../../common/configuration'
+
+// Mainnet Addresses
+export const USDA = networkConfig['31337'].tokens.USDA as string
+export const StUSD = networkConfig['31337'].tokens.stUSD as string
+export const USDA_USD_PRICE_FEED = networkConfig['31337'].chainlinkFeeds.USDC as string // we use USDC feed as USDA/USD is not available in chainlink, and USDA/USDC is always 1 through the transmuter
```
pmckelvy1, line 7: ok this is a critical piece of the pricing puzzle so i think its worth justifying this with a more in-depth explanation.
can you link the transmuter docs and code?
is USDC **ALWAYS instantly and atomically exchangeable 1:1** for USDA?
are there any failure modes?

pmckelvy1 (changes requested):

common/configuration.ts
nlecoufl: If you look at the token symbol https://basescan.org/address/0x0000206329b97DB379d5E1Bf586BbDB969C63274#readProxyContract#F17, it's `USDA`. We'll check to update it, `agUSD` was used in the first version of the token and was subsequently renamed to `USDA`.

contracts/plugins/assets/angle/USDAFiatCollateral.sol
nlecoufl: Got it, but wondering if this is really needed for `stUSD`. View functions like `totalAssets` or `convertToAssets` already includes accruals (see https://github.com/AngleProtocol/angle-transmuter/blob/main/contracts/savings/Savings.sol#L109). So it is possible to compute up-to-date prices without having to trigger a state update, isn't it?

scripts/addresses/1-tmp-assets-collateral.json
nlecoufl: Yep reverting, it was deployed on a local mainnet fork.

test/plugins/individual-collateral/angle/constants.ts
nlecoufl: Please find here the Transmuter [code](https://github.com/AngleProtocol/angle-transmuter) and [docs](https://docs.angle.money/transmuter/transmuter).
USDA is in the current settings exchangeable 1:1 instantly and atomically with USDC.
As you point out, there are some failure modes (and this is where the value of the Transmuter lies), especially if USDC was to depeg. Transmuter was designed to incorporate anti-depeg mechanisms, and automatically fallback into a preservation mode if one of the assets in its backing was to depeg. As currently all assets for which this monitoring is activated are USDC derivatives (namely USDC and steakUSDC on Morpho) then, only a USDC depeg could cause the 1:1 exchangeability to break between the 2 assets.
There are some cases where the protocol could have its liquidity buffer in USDC depleted, but in this case, the system supports a permissionless rebalancing feature allowing anyone to refill the USDC reserves of the protocol and enable 1:1 USDA->USDC conversion.

common/configuration.ts
pmckelvy1: ok i think you need to reach out to etherscan to get them to refresh it

contracts/plugins/assets/angle/USDAFiatCollateral.sol
pmckelvy1: yes but there are state updates performed in the collateral plugin, so we should match that with any necessary state updates in the collateral itself.

test/plugins/individual-collateral/angle/constants.ts
pmckelvy1: i think having the liquidity buffer depleted is a concern. there's really no guarantees on when it would be refilled, right? and at that point, users would need to go to the market to exchange?

nlecoufl: In the current setup with the public harvester contract, and the liquidity of the collateral assets, there is in fact a guarantee that this liquidity buffer will be refilled.
While we mentioned the liquidity buffer in USDC, with the transmuter, users can burn their USDA for any of the other assets in the backing (which is not the case by the way with the majority of other stablecoin protocols), and so before selling on the market, they can burn for like steakUSDC and USDM. They may also choose to redeem their USDA for a portion of all the assets in the backing depending on the proportion that they represent.
In any case, users are never locked in Angle and can always exit using the system's primary minting mechanism

pmckelvy1:
leaving this pr open for now. need an stUSDA oracle in order to complete the plugin.

PR #1220, opened 2024-10-18
Unpriced collateral

tbrent:


contracts/facade/factories/CollateralFactory.sol
```diff
@@ -2,14 +2,18 @@
...
- * @title DemurrageCollateralFactory
+ * @title CollateralFactory
  */
-contract DemurrageCollateralFactory {
+contract CollateralFactory {
     event DemurrageCollateralDeployed(address indexed collateral);
+    event UnpricedCollateralDeployed(address indexed collateral);
 
-    mapping(address coll => uint192 feePerSecond) public demurrageDeployments;
+    mapping(address coll => uint192 feePerSecond) public demurrageDeployments; // not unique by erc20
```
akshatmittal, line 14: `address erc20` => `uint256/uint192 fee` => `address collateral` is much better to query and track.

tbrent: there is not just 1 collateral address possible, per fee tier

akshatmittal: Sorry I meant for unpriced demurrage collateral. There aren't any specific parameters for those, I don't think right?

tbrent: that's 2 lines below and works the way you want it to

pmckelvy1: well it's saving by address of the deployed collateral. maybe this comment should go in the other pr, but if we save a mapping like akshat suggested (and error on duplicates), then the fe can easily query the contract based on the erc20 addy. otherwise we're requiring a subgraph dependency

tbrent:
tabling plans to use unpriced collateral for now

PR #1223, opened 2024-10-30
OETH L2 Base plugin

clement-ux:
- OETH plugin and tests for L2 Base.

PR #1226, opened 2024-11-06
Oeth L2 base plugin

tbrent:
workaround for github secrets stuff from #1223


contracts/plugins/assets/origin/OETHCollateralL2Base.sol
```diff
@@ -0,0 +1,87 @@
...
+ * tar = ETH
+ * UoA = USD
+ */
+contract OETHCollateralL2Base is ERC4626FiatCollateral {
+    using OracleLib for AggregatorV3Interface;
+    using FixLib for uint192;
+
+    IMorphoChainlinkOracleV2 public immutable targetPerTokChainlinkFeed; // {tar/token}
+
+    AggregatorV3Interface public immutable uoaPerTargetChainlinkFeed; // {UoA/tar}
```
julianmrodri, line 27: why not use the default chainlinkFeed for this one? I think this is what we did for RETH for example.

test/plugins/individual-collateral/origin/OETHCollateralL2Base.test.ts
```diff
@@ -0,0 +1,258 @@
...
+import { mintWSUPEROETHB } from './helpers'
+import { expect } from 'chai'
+import { ethers } from 'hardhat'
+import { ContractFactory, BigNumber, BigNumberish } from 'ethers'
+import {
+    ERC20Mock,
+    MockV3Aggregator,
+    MockV3Aggregator__factory,
+    TestICollateral,
+    IWSuperOETHb,
```
julianmrodri: missing will replace with `IERC4626` as no special logic is needed

contracts/plugins/assets/origin/OETHCollateralL2Base.sol
```diff
@@ -0,0 +1,87 @@
...
+        uint192 p = uoaPerTar.mul(targetPerTok);
+        uint192 err = p.mul(oracleError, CEIL);
+
+        high = p + err;
+        low = p - err;
+        // assert(low <= high); obviously true just by inspection
+
+        // {tar/ref} = {tar/tok} / {ref/tok} Get current market peg
+        // ETH/superOETHb = ETH/wsuperOETHb / superOETHb/wsuperOETHb
+        pegPrice = FIX_ONE;
```
julianmrodri, line 85: shouldnt we do here `pegPrice = targetPerTok.div(underlyingRefPerTok();` ? I believe should provide a similar result and it's already implemented in some plugins

test/plugins/individual-collateral/origin/OETHCollateralL2Base.test.ts
```diff
@@ -0,0 +1,268 @@
...
+import { mintWSUPEROETHB } from './helpers'
+import { expect } from 'chai'
+import { ethers } from 'hardhat'
+import { ContractFactory, BigNumber, BigNumberish } from 'ethers'
+import {
+  ERC20Mock,
+  MockV3Aggregator,
+  MockV3Aggregator__factory,
+  TestICollateral,
+  IWSuperOETHb,
```
julianmrodri: file missing, replaced with `IERC4626` as no special logic is needed

julianmrodri (commented):
Overall looks ok, had to do some tweaks to get tests running.. Have a few questions on a few things in the plugin code. Also we definitely want a `README` like we have for other plugins (in the `plugin/assets/origin` folder)

contracts/plugins/assets/origin/OETHCollateralL2Base.sol
tbrent: this was what we had previously, but it turns out they are using the same underlying information. the price shown in the `targetPerTok` oracle is the same as the `underlyingRefPerTok` at all times, so I suggested this change as a gas-saver

can't set to 0 since the inherited `refresh()` function checks the value

contracts/plugins/assets/origin/OETHCollateralL2Base.sol
tbrent: no reason in particular, just didn't want to refactor the submitted code. I agree it's dissimilar to how we've done it in other cases but it seems fine

tbrent:
test seem to be failing due to something about the chain not resetting correctly between tests, I think? not sure why else the addition a new testfile would have broken the old stable collateral

julianmrodri:
> test seem to be failing due to something about the chain not resetting correctly between tests, I think? not sure why else the addition a new testfile would have broken the old stable collateral

yeah probably need to adjust the precision in those tests which rely on timestamp, which i assume now its different  its not a big deal and unrelated to the code changes per se.

julianmrodri:
@clement-ux whats the idea of using the `MorphoChainlinkFeed` here. Because based on current settings it only takes information from the underlying vault, which we can do without going through that contract.  Is there more configuration in the oracle expected which would justify keeping it as a separate contract?

PR #1230, opened 2024-11-25
Fixes and new volatile plugins and addresses

julianmrodri:
* New plugins for WELL, cbBTC and DEGEN
* Redeployed previous volatile plugins (MOG and WETH_AERO)
* implemented fixes in docs and comments as suggested by Trust
* Set peg price = 0 for stable plugins


pmckelvy1 (approved):
lgtm

PR #1236, opened 2025-01-29
MidasRWA: mBTC plugin

julianmrodri:


julianmrodri:
@dmytro-horbatenko some comments regarding the PR:

* I suggest having different contracts for the mBTC one (NonFiat) and for the one that is pegged to USD (Fiat) instead of having logic regarding target name in the plugin. 

* Tests should be in the common format as all other 99% plugin tests using the generic suite `collateralTests.ts` , and adding specific tests if required. The tests in the PR need to be refactored to that format.

* There are some tests failing as well.

Fixing this is not a guarantee we will proceed with the plugin as there are other requirements to be considered in terms of liquidity, oracles, fees, restrictions which all seem to apply to Midas and would require further analysis.

pmckelvy1:
see https://github.com/reserve-protocol/protocol/pull/1238

PR #1238, opened 2025-03-17
MidasRWA: collateral plugins for mBTC and mTBILL

dmytro-horbatenko:


pmckelvy1:
@dmytro-horbatenko it is our understanding that the mBTC and mTBILL price oracles are fully managed by the Midas team.  The centralized nature of these critical oracles means that we cannot recommend these plugins as safe for integration into RTokens.
If you are able to secure more a decentralized oracle for these prices, we can reconsider this PR.

PR #1241, opened 2025-04-10
OETH Plugin -Mainnet

julianmrodri:


test/plugins/individual-collateral/origin/OETHCollateral.test.ts
```diff
@@ -0,0 +1,259 @@
...
+  itChecksPriceChanges: it,
+  itChecksNonZeroDefaultThreshold: it.skip,
+  itHasRevenueHiding: it,
+  resetFork: getResetFork(FORK_BLOCK),
+  collateralName: 'OETHCollateral',
+  chainlinkDefaultAnswer: defaultAnswers.targetPerRefChainlinkFeed,
+  itIsPricedByPeg: true,
+  itHasOracleRefPerTok: false,
+  targetNetwork: 'mainnet',
+  toleranceDivisor: bn('1e2'),
```
tbrent, line 256: can we increase this toleranceDivisor? most other collateral use 1e8 or 1e9. The only case we have of 1e2 is aerodrome plugins or LP tokens collateral because those are harder to measure, but we don't have that going on here

julianmrodri: @clement-ux tagging you as this is a new branch/pr

julianmrodri: Seems like `1e3` is the highest we can go. Should we dig deeper? ITs in the "increase" price functions that the comparison fails maybe need to be adjusted.

clement-ux: @julianmrodri Sorry I think I didn't get it. 
Are you waiting for me to modify `toleranceDivisor`? Or this is a conversation you have internally? 
Because I don't see any failing test on the CI.

julianmrodri: Let me check with Taylor, but it is true we should probably be able to increase it in the test. But if we do that the test fails which indicates to me we can improve the test somewhere. I havent checked in detail but price comparison tests are impacted so probably the expected price we calculate is off or we are missing updating one of the oracles or something in those lines. If you go to `1e4` you will see failures

PR #1254, opened 2025-08-26
4.2.0 - Trusted fillers

julianmrodri:
* Add trusted filler support for Dutch Trades 
* Adds tests, docs, and deployment setup


contracts/plugins/trading/DutchTrade.sol
```diff
@@ -283,7 +321,15 @@ contract DutchTrade is ITrade, Versioned {
...
-        require(bidder != address(0) || block.timestamp > endTime, "auction not over");
+        require(
+            bidder != address(0) ||
+                address(activeTrustedFill) != address(0) ||
+                block.timestamp > endTime,
+            "auction not over"
+        );
+
+        // Close any active trusted fill and reclaim tokens
+        _closeTrustedFill();
```
tbrent, line 332: We can't confirm the buyer transferred any tokens through the filler, right? Anyone can call `trading.settleTrade(..)` to deny the auction. I think there aren't any other consequences other than denying it, but still, we need a slightly different design here (unless I'm missing something).

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -274,6 +281,37 @@ contract DutchTrade is ITrade, Versioned {
         assert(status == TradeStatus.CLOSED);
     }
 
+    /// Create a trusted fill as an alternative to direct bidding
+    /// @param targetFiller The trusted filler implementation to use (e.g., CowSwapFiller)
+    /// @param deploymentSalt Unique salt for deterministic deployment
+    /// @return filler The deployed trusted filler contract
+    function createTrustedFill(address targetFiller, bytes32 deploymentSalt)
```
tbrent, line 288: I think this can be called for free to DoS `bid()` calls

contracts/plugins/trading/DutchTrade.sol
julianmrodri: Looking at it now, I think the `require` here is wrong and we should probably enforce reaching the auction end time in the case of Trusted fillers.

contracts/plugins/trading/DutchTrade.sol
julianmrodri: Yes. `bid()` would not work after this is called because tokens leave the DutchTrade.

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -274,6 +281,37 @@ contract DutchTrade is ITrade, Versioned {
...
+        external
+        returns (IBaseTrustedFiller filler)
+    {
+        assert(status == TradeStatus.OPEN);
+
+        // Get trusted filler registry
+        ITrustedFillerRegistry registry = IExtendedBroker(address(broker)).trustedFillerRegistry();
+        bool enabled = IExtendedBroker(address(broker)).trustedFillerEnabled();
```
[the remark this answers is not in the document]
julianmrodri: You can only enable/disable once is set.

contracts/plugins/trading/DutchTrade.sol
julianmrodri: Added the `closeTrustedFiller` modifier as a pre-hook to ensure `bid` can be called.

contracts/plugins/trading/DutchTrade.sol
julianmrodri: Modified `settleTrade` to only settle the ongoing auction if we have the `buy` tokens we require from the filler.

julianmrodri:
> Implementation LGTM
> 
> There's a weird thing where you can call `bid()` or `bidWithCallback()` after the open trusted filler has been filled. I don't think you can do anything damaging with it yet, just noting it's weird

Doesnt the same happen on Folios?

contracts/plugins/trading/DutchTrade.sol
```diff
@@ -194,17 +210,17 @@ contract DutchTrade is ITrade, Versioned {
             FLOOR
         );
         uint192 _bestPrice = prices.sellHigh.div(prices.buyLow, CEIL); // no additional slippage
-        assert(_worstPrice <= _bestPrice);
+        require(_worstPrice <= _bestPrice, "invalid pricing");
```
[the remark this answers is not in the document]
tbrent: require there imo

julianmrodri:
LGTM with the contract size changes.

PR #1263, opened 2025-10-08
Ether.fi weETH plugin

julianmrodri:
* Plugin for `weETH`
* Includes tests, deployment and verification scripts.


tbrent (approved):
Nice job, couldn't find anything. Only thing I can think of is we could add a small section to the `README.md` noting the decision not to monitor the eETH<->ETH exchange rate, as well as a comment addressing the decision not to handle any extra reward tokens right now.

PR #1267, opened 2025-10-30
AssetRegistry RTokenAsset

tbrent:
ended up preferring bumping the solidity version to 0.8.28 over inverting all require statements in AssetRegistry to switch to custom errors. That's why the Files changed is so high. Keeps more of our tests untouched though


contracts/p1/AssetRegistry.sol
akshatmittal: Do we not need to special case the RTokenAsset here so it's not checked against the registry?

PR #1291, opened 2026-06-08
MorphoV2 plugins 

julianmrodri:
- Creates `MorphoV2FiatCollaterla.sol` extending the existing  `MetaMorphoFiatCollateral` plugin for 8 mainnet Morpho Vault V2 vaults (steakUSDC Prime, Paypal PYUSD, Gauntlet USDC Frontier/Prime, Galaxy USDC/USDT Quality, Steakhouse USDT Prime,
  Sky USDT Savings) 
- Tests extending the meta-morpho fork suite with all 8 vaults
- Deploy and verification scripts

## Summary by CodeRabbit

* **New Features**
  * Added support for Morpho Vault V2 collateral across supported Ethereum deployments.
  * Added safeguards for restricted vault gates, collateral status transitions, and permissionless Merkl reward claiming.
  * Added MORPHO pricing support using Chainlink and Uniswap TWAP data.
  * Added deployment and verification support for supported Morpho Vault V2 assets.

* **Bug Fixes**
  * Corrected MORPHO token addresses while preserving legacy MORPHO handling for existing Aave V2 collateral.

* **Documentation**
  * Updated Morpho reward, Vault V2, and collateral guidance with current links and deployment details.


contracts/plugins/assets/meta-morpho/README.md
```diff
@@ -39,6 +39,32 @@ It is important to note that in the case of Rtokens, rewards will need to be cla
...
+| Name                   | Symbol       | Address                                      | Asset |
+| ---------------------- | ------------ | -------------------------------------------- | ----- |
+| Steakhouse Prime USDC  | steakUSDC    | `0xbeef088055857739C12CD3765F20b7679Def0f51` | USDC  |
+| Sentora PYUSD Main     | senPYUSD     | `0xC21b08C16458202593D4D9B26b9984Ee67b38BbD` | PYUSD |
+| Gauntlet USDC Frontier | gtusdcf      | `0x9a1D6bd5b8642C41F25e0958129B85f8E1176F3e` | USDC  |
+| Steakhouse Prime USDT  | steakUSDT    | `0xbeef003C68896c7D2c3c60d363e8d71a49Ab2bf9` | USDT  |
+| Galaxy USDT Quality    | gUSDTq       | `0x71ffB6a81786eC285D429d531Cf655107B9D878d` | USDT  |
+| Gauntlet USDC Prime    | gtusdcp      | `0x8c106EEDAd96553e64287A5A6839c3Cc78afA3D0` | USDC  |
+| Galaxy USDC Quality    | gUSDCq       | `0x91600E31fBeDc72433d4a57F16639cfe661Be7d8` | USDC  |
+| Sky.money USDT Savings | skyMoneyUsdt | `0x23f5E9c35820f4baB695Ac1F19c203cC3f8e1e11` | USDT  |
```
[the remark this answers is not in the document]
julianmrodri: This is not a problem, one talks about the vault in general not the identifier in our configuration

tbrent (dismissed):
Looks good. Only thing to mention, maybe intended

> README symbols for Sentora and Sky do not exactly match on-chain metadata. Sentora returns senPYUSDPRIMEv2, not senPYUSD; Sky returns skyMoneyUsdtSavings, not skyMoneyUsdt.

julianmrodri:
> Looks good. Only thing to mention, maybe intended
> 
> > README symbols for Sentora and Sky do not exactly match on-chain metadata. Sentora returns senPYUSDPRIMEv2, not senPYUSD; Sky returns skyMoneyUsdtSavings, not skyMoneyUsdt.

No.. will put the real ones int he README I thought I had that covered

tbrent (dismissed):

akshatmittal (commented):
Detailed review found three issues in the production MorphoAsset and verification paths. Inline comments include impact and remediation. The Vault V2 gate checks otherwise match Morpho's semantics, and the eight fork suites pass.

contracts/plugins/assets/meta-morpho/MorphoAsset.sol
```diff
@@ -0,0 +1,151 @@
...
+ *   - {UoA/quoteTok} comes from `chainlinkFeed` (e.g. ETH/USD)
+ *   - {quoteTok/tok} comes from an arithmetic-mean-tick TWAP over `uniswapV3Pool`
+ *
+ * WARNING -- operational requirements for a safe deployment:
+ *   1. The pool's `observationCardinality` MUST be large enough to retain `twapWindow` seconds of
+ *      observations, or `observe()` reverts ("OLD") and this Asset becomes unpriced. Cardinality
+ *      is grown permissionlessly via `increaseObservationCardinalityNext()` and never shrinks;
+ *      the constructor probes the window once, but a pool that is traded in many consecutive
+ *      blocks can still evict history later. Size cardinality with headroom.
+ *   2. A TWAP is only as manipulation-resistant as the pool is deep. `maxTradeVolume` -- not the
```
akshatmittal: **P1 — `maxTradeVolume` does not bound true-value exposure when the TWAP is the attacked input.** `TradeLib.maxTradeSize` converts the $10k limit into MORPHO by dividing by this same `sellHigh`. If true price `P` is pushed to `kP`, the lot grows to approximately `M / (1.05 * kP)`, with true value `M / (1.05 * k)`, while minimum proceeds remain around `M * 0.95 / 1.05` before slippage and buy-side error. At a 50% TWAP, about $19k of MORPHO can therefore be sold against a roughly $9k floor; at lower prices exposure grows until the entire held balance is the cap. The stated protection fails precisely under the manipulation scenario it is intended to contain. Since anyone can launch the auction and the pool is documented as thin, please remove the production `MorphoAsset` wiring until sizing uses an independent price or an oracle-independent token/aggregate exposure cap. This need not block the Vault V2 collateral plugins.

scripts/verification/collateral-plugins/verify_legacy_morpho.ts
```diff
@@ -37,7 +39,7 @@ async function main() {
         rewardsDistributor: networkConfig[chainId].MORPHO_REWARDS_DISTRIBUTOR!,
         underlyingERC20: networkConfig[chainId].tokens.USDT!,
         poolToken: networkConfig[chainId].tokens.aUSDT!,
-        rewardToken: networkConfig[chainId].tokens.MORPHO!,
+        rewardToken: LEGACY_MORPHO,
```
akshatmittal, line 45: **P1 — `LEGACY_MORPHO` is not imported or defined.** This produces a deterministic `ReferenceError` when the verifier reaches this constructor argument (`eslint --rule no-undef:error` reports it at 42:22). `verify_etherscan.ts` invokes this legacy verifier before both new V2/Morpho verification scripts, so the full verification workflow stops here. Please import the existing `LEGACY_MORPHO` constant or define the legacy token in a shared configuration source.

scripts/deployment/phase2-assets/assets/deploy_morpho.ts
```diff
@@ -0,0 +1,131 @@
...
+  const pool = await hre.ethers.getContractAt('IUniswapV3Pool', MORPHO_WETH_POOL_030)
+  const slot0 = await pool.slot0()
+  const requiredCardinality = Math.ceil(MORPHO_TWAP_WINDOW / BLOCK_TIME)
+  console.log(`
+    Uniswap V3 pool:          ${MORPHO_WETH_POOL_030}
+    observationCardinality:   ${slot0.observationCardinality}
+    ...Next:                  ${slot0.observationCardinalityNext}
+    required for ${MORPHO_TWAP_WINDOW}s window: ${requiredCardinality}`)
+
+  if (slot0.observationCardinalityNext < requiredCardinality) {
```
akshatmittal: **P2 — this known oracle-liveness prerequisite should fail deployment, not only warn.** The passing deployment CI recorded `observationCardinality = observationCardinalityNext = 51` versus 150 requested. As the warning itself notes, enough swaps can overwrite the retained 30-minute history, make `observe(1800)` revert `OLD`, and eventually leave the asset unpriced so rewards cannot be auctioned. Please grow `observationCardinalityNext` with headroom before deployment and abort while it remains undersized. Also note that 150 observations span only 149 twelve-second intervals; the strict minimum for 1,800 seconds is 151, though the suggested 300 is preferable.

tbrent (approved):
We need to manually call `increaseObservationCardinalityNext()` I think to increase the cardinality to 120-150 minimally, or 250 ideally, I think. Otherwise looks good

tbrent (changes requested):
actually, we should probably increase priceError from 5% to 10% as well, just given pool is low liquidity and we're measuring a TWAP

contracts/plugins/assets/meta-morpho/MorphoAsset.sol
julianmrodri: So we cannot deploy this? Blocked until oracle available?

tbrent: I think we can deploy, we just have to be okay with the risks. Akshat's comment is codex, not his own claim

julianmrodri: Ok buyt the risk is there. The pool only holds about $62k, which is cheap for someone to push around.

```
The problem is the protocol sizes each auction as `maxTradeVolume ÷ our reported price`. 

So if an attacker halves the price we report, we hand over twice as many MORPHO — while the minimum we're required to accept stays fixed at ~$8,100 no matter how far they push it.

That means our "$10k per auction" cap doesn't actually cap anything real: the only true limit is the entire MORPHO balance sitting in the Backing Manager. 

Every other plugin is safe here because their prices come from Chainlink, which a bidder can't move — we're the first to put a thin AMM in that denominator.
```

julianmrodri: Worth discussing today because there are also no `MORPHO` rewards on any of these vaults.

akshatmittal: @tbrent It's Codex yes, but it's my claim too haha.

While I think the liquidity is very low and can be manipulated, I think for a reward only thing this is ok. That said, we should NOT deploy this right now given that we don't need it at all. We do have other tokens that have MORPHO rewards tho, so that might be a reason to deploy anyway.

tbrent (approved):

contracts/plugins/assets/meta-morpho/MorphoAsset.sol
julianmrodri: We will not be claiming or trading Morpho rewards

scripts/deployment/phase2-assets/assets/deploy_morpho.ts
julianmrodri: fixed

scripts/verification/collateral-plugins/verify_legacy_morpho.ts
julianmrodri: fixed

